PR #11, opened 2015-05-06
BSOD Bad_pool bug Changes 

dimov-cz:
1) I used  0.6.0 for while and had BSOD very often.
2) Than i patched according to advices from based on advice from https://code.google.com/p/dokan/issues/detail?id=229 adn with this i had no BSOD for almost year.
3) Now after upgrade to this version 0.7.1(because this one have subscribed 64bit drivers ) I got BSOD today again. It typically happen if u use intensive parallel reading/writing.

### DIFF
--- a/sys/create.c
+++ b/sys/create.c
@@ -149,10 +149,11 @@ DokanGetFCB(
 		ExFreePool(FileName);
 	}
 
+	InterlockedIncrement(&fcb->FileCount);
+
 	ExReleaseResourceLite(&Vcb->Resource);
 	KeLeaveCriticalRegion();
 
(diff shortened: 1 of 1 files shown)


dimov-cz:
Note: This patch already is included in current DokanX/master.

Liryna:
Oh yes well found ! it looks like this dokanx feature has been missed.
We will check about it and probably make a signed release during the next week :)

Maxhy:
Thanks for this good PR. Merged.

Maxhy:
@dimov-cz I will hijack a bit this thread but I didn't found any other way to contact you.
Could you send me an email so we can discuss about SSH File System future directly? Thanks.

PR #13, opened 2015-05-11
Fix DbgPrint message for CreateDisposition

suy:
It was printing the number in hexadecimal, but appending 08X at the end.
Fixed to add 0x at the beginning, which probably was the intended
output.


Maxhy:
Good catch, thanks!

PR #14, opened 2015-05-11
Fix DokanUnmount bug

suy:
As reported in the mailing list, DokanUnmount called
DokanRemoveMountPoint with an extra character at the end. Fixed it, and
mention DokanRemoveMountPoint in the readme.txt, since is an exported
function in the public header and is necessary if you mount the file
system in a path, not a drive letter.

### DIFF
--- a/dokan/mount.c
+++ b/dokan/mount.c
@@ -237,7 +237,7 @@ BOOL DOKANAPI
 DokanUnmount(
 	WCHAR	DriveLetter)
 {
-	WCHAR mountPoint[] = L"M:\\";
+	WCHAR mountPoint[] = L"M:";
 	mountPoint[0] = DriveLetter;
 	return DokanRemoveMountPoint(mountPoint);
 }
--- a/dokan/readme.txt
+++ b/dokan/readme.txt
@@ -313,10 +313,11 @@ otherwise, the following error code is returned.
 Unmounting
 ==========
 
-File system can be unmounted by calling the function DokanUnmount.  In
-most cases when the programs or shells use the file system hang,
-unmount operation will solve the problems by bringing the system to
-the previous state when the file system is not mounted.
+File system can be unmounted by calling the function DokanUnmount or
+DokanRemoveMountPoint.  In most cases when the programs or shells use
+the file system hang, unmount operation will solve the problems by
+bringing the system to the previous state when the file system is not
+mounted.
 
 User may use DokanCtl to unmount file system like this:
    > dokanctl.exe /u DriveLetter


suy:
Sorry if you received a notification for this twice. I force-pushed the commit just to amend the message and include the link to the discussion.

suy:
I think this is because the string passed to tell where you mount it has to be the same when you unmount. Since the mirror.exe example mounts in a drive letter, the string that has to be passed when unmounting has to be one single letter as well. That's why I added it to the readme, because I think you should probably use the long version. That way you always have the same string for sure.

The patch changed DokanUnmount, though. The mirror uses the other one, so I think should not be affected.

Liryna:
You are right, I made more test.
Mount "M://" generate a DOKAN_MOUNT_POINT_ERROR but not "M:".

MaMic:
Ah, I see, sorry for my confusion.
So one has to unmount by using the same string as used in mount ... ok I think I understand now.
So DokanUnmount is actually pretty useless, but must persist because of compatibility ... and because of that I think it should be left as it was the original way.
But in short ... it's better to always use RemoveMountPoint, right?

Liryna:
Using Unmount or RemoveMountPoint make no difference.
But I agree with you that Unmount have no really meaning other that simply propose to Unmount by giving a letter and not a path.

MaMic:
Yes, all right, but I feel that somehow for compatibility and symmetry reasons, the single character (drive letter) unmount should be converted to use a mount point of the form L"M:\\", as it was before this commit.

My experiments showed that in any case exactly the same string has to be used in RemoveMountPoint, as it was specified during the mount operation. Even upper - and lowercase have influence, which is also ignored currently.

In mirror.c a single character is accepted in the command line arguments, but is converted to the form L"M:", so yes, if for this client the Unmount('M') leads to RemoveMountPoint(L"M:"), that seems to work, but what if the client just forwards the 'M' as L"M", without the ':', the Unmount('M') would fail. So each try to simplify things with such kind of helper do confuse more and more, because not all cases or intentions of the user can be covered.

Another hint for keeping backward compatibility can be found here e.g. in dokan.c line 191, where older versions used a single letter mount point that is altered to the same form L"M:\\".

So for maximum backward compatibility and simplified user experience I would suggest to convert each single character input (whether in mount, unmount or anywhere else) by appending a colon sign and backslash and take the upper case string of that as the new mount point of the form L"M:\\"

Liryna:
I agree with you!

Your proposal seems to be fine!
Appending a colon sign and backslash and upper case, would make it easier
We would keep the Unmount for compatibility and call RemoveMountPoint directly with this changes.

This should fix all misunderstanding.
(If no one see a potential wrong behaviour)

PR #19, opened 2015-05-28
Amend contradictory DeleteFile comment in header

suy:
The same header says 'Delete on when "cleanup" is called' on the DokanFileInfo
definition, and "you must delete the file in Cleanup". Also, this contradicts
the mirror application behaviour, where the file/directory deletion is done on
Cleanup, not CloseFile.


Maxhy:
Indeed, much better explanation.

PR #21, opened 2015-06-01
Include windows.h in the main dokan header file

sherter:
If someone forgets to #include windows.h before dokan.h is included, weird
errors appear. It is good practice for header files to be self-contained.


Maxhy:
Indeed this is good practice here, thanks.

PR #23, opened 2015-06-09
Exported function names for x86/x64 library

MaMic:
Tried the current v0.7.3-beta release and could not bind to the library because of mangling/decorated function names in the x86 version of the library. Hope that my changes ( - works on my machine ;-) - ) correspond to the usual/common approach to handle this kind of problem.

Had to change the additional dependencies in the project files for control and mirror in order to build complete solution on my machine, is this only a problem on my site?

### DIFF
--- a/dokan/dokan.h
+++ b/dokan/dokan.h
@@ -27,7 +27,7 @@ with this program. If not, see <http://www.gnu.org/licenses/>.
 #define DOKAN_DRIVER_NAME	L"dokan.sys"
 
 #ifdef _EXPORTING
-  #define DOKANAPI __declspec(dllexport) __stdcall
+  #define DOKANAPI /*__declspec(dllexport)*/ __stdcall // exports defined in dokan.def
 #else
   #define DOKANAPI __declspec(dllimport) __stdcall
 #endif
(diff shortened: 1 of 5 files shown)


Maxhy:
I noticed the behavior too but I don't get it.
def file should be optional, both ways should end to the same result and it is on x64 build but not x86. I believe you didn't find out way neither as you moved to def file?

MaMic:
The both variants __declspec and .def file can lead to the same result, but it does not need to. By using the .def file you can export the mangling name, plain name or even no name. Using __declspec is a simple way to even export whole complex c++ classes, but mangling names are nearly necessary then (what is no problem, as long as you just link the .lib using the same compiler that was used building the library).

__declspec(dllexport) __stdcall under x86 always leads to mangling names, even with export "C", the reason for that is the calling convention. By using __cdecl, we would have the plain name. This seems also to be the reason for Microsoft to do it that way, they use __stdcall calling conventions and .def files for the Win32 API.

So what is the difference in the x64 build? Quite simple, __sdtcall is accepted by the x64 compiler, but ignored! Here __fastcall is always used, what again turns to another behavior for our case.

At least as far as I understand, but must not necessarily be true, since I'm not really expert in the field ;-)

Maxhy:
Ok thanks for the explanation this makes the behavior clearer.
I merged your PR, thanks for the changes!

PR #39, opened 2015-08-09
MoveFileEx fix

rustyx:
This is a fix for CreateFile with a SL_OPEN_TARGET_DIRECTORY flag, which should open the containing directory instead of the file itself.
This flag is used in the MoveFileEx WinAPI with the MOVEFILE_REPLACE_EXISTING flag, which opens the target file's containing directory before deleting the target file. Without the fix this API fails with "permission denied" because the target file is opened and cannot be deleted.
Now the git push command works via dokan!

### DIFF
--- a/dokan/create.c
+++ b/dokan/create.c
@@ -87,6 +87,22 @@ DispatchCreate(
 		//DbgPrint("FILE_DIRECTORY_FILE\n");
 		directoryRequested = TRUE;
 	}
+	else {
+	    if (EventContext->Flags & SL_OPEN_TARGET_DIRECTORY) {
+		DbgPrint("SL_OPEN_TARGET_DIRECTORY specified\n");
+		// strip the last section of the file path
+		WCHAR* lastP = NULL;
+		for (WCHAR* p = EventContext->Operation.Create.FileName; *p; p++) {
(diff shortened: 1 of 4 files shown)


Liryna:
Thank you for the contribution !

Well found ! I have never faced this issue since I never did such operation :yum: 

You PR seems to be great but is their a reason for removing this ?
https://github.com/dokan-dev/dokany/pull/39/files#diff-a8f32b6fd8169cac1606e8d8b0f497c1L184
Now we have `warning C4800: 'unsigned long' : forcing value to bool 'true' or 'false' (performance warning) ` during the build :cry:

rustyx:
Hi,
Ah yes this was another bug. The code was:
```
	bool read=(Flags & ACCESS_READ) == 1;
	bool write=(Flags & ACCESS_WRITE) == 1;
```
\- this would never work as neither ACCESS_READ nor ACCESS_WRITE equal 1 !
So both flags were always false.

To fix the warning you could try:
```
	bool read=(Flags & ACCESS_READ) != 0;
	bool write=(Flags & ACCESS_WRITE) != 0;
```

Liryna:
Perfect !
Thanks for the informations :) !

PR #41, opened 2015-08-11
Heavily redone comments

arekbulski:
Some statements were ambiguous and not really meaningful. I tried to be as accurate in representing what the comments would supposedly mean. 

I also worked heavily on the wiki, which is not part of this pull request.


Liryna:
Thank you again ! :trophy:

PR #55, opened 2015-09-06
Irp pending timeout issue

marinkobabic:
* The timeout per device has been implemented. The DokanOptions has a new property Timeout (Milliseconds).
* Prevent thread termination if not enough resources
* Fix BSOD in security

@Liryna 
Please create a .Net Method ResetTimeout. The exported method exists since a longer time https://github.com/dokan-dev/dokany/blob/master/dokan/dokan.def#L13. This method can be executed inside of the callbacks, to extend the timeout of the current request.

Please add the new DokanOption.Timeout

### DIFF
--- a/dokan.sln
+++ b/dokan.sln
@@ -44,7 +44,6 @@ Global
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
 		{C0269F86-24E0-40DD-B5C8-29CDFDFF2AFB}.Debug|Win32.ActiveCfg = Win8.1 Debug|Win32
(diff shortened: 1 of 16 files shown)


Maxhy:
At first look/test this is good!
You really did an amazing quick job, thanks for this PR. I'm gonna merge it but additional tests would be preferable to be sure it breaks nothing. Lets consider *master* as unstable for now ;).

Liryna:
Very very nice PR @marinkobabic TY :heart:  !
I updated the TODO and will update the .NET wrapper as you requested for the next release :) !

Liryna:
@marinkobabic 
Timeout has been added https://github.com/dokan-dev/dokan-dotnet/blob/master/DokanNet/Native/DOKAN_OPTIONS.cs#L15
ResetTimeout was already there https://github.com/dokan-dev/dokan-dotnet/blob/master/DokanNet/Native/NativeMethods.cs#L25

PR #56, opened 2015-09-07
Exception handling

marinkobabic:
Have now implemented the central error handling. This helps to make the driver stable and to keep some standards for each Irp. Next step would be to create IrpContext like fastfat and also to extend the exception handler.


Maxhy:
Much better architecture, thanks again @marinkobabic ;)

PR #64, opened 2015-09-24
Attempt to fix file mapping. Added a ported fsx utility

andreyloktev:
Added the fsx utility. I did not test all options. To try to reproduce a file mapping problem on the current release dokan version a command "fsx.exe -N 1000 -P C:\fsxtest K:\fsxtestfile" should be used. Where -N: number of operations, -P: path to fsx.good and fsx.log files and K:\ fsxtest- a testing file on a dokan drive


Liryna:
Hi @andreyloktev ,
Thank you for the pull request !

I will try to review it when I get free time but at first look it look really nice :+1: 

About fsx, do you know/have used some other commands ? (even a list would be nice :D)
I will try to add them as unit test in appveyor.

andreyloktev:
Hello @Liryna,
fsx includes only read, write, mapread, mapwrite and truncate operations. Than less options are used then more operations will be done :smile: . For example, -R, -W disable map read and map write.
Although there are usefull options for debugging:
1) fsx calls dowrite(), domapwrite(), ... in a random way. It initializes a random generator and prints a seed number before to begin testing. If error occurs you will be able to reproduce it using -S to specify the printed seed number.
2) -i and -I - enable an interactive mode to do operations step by step to see a dokan log
3) -b to specify a beginning opration number

Option -N is used to specify the number operations to do. If it is not used fsx will run infinitely. I think 2000000 operations is enougth :smile: 

Later I would like to port and to add a fstorture utility. It tests file system operations: cerate, delete, raname, stat, truncate a file;  create, delete, rename a directory; creation a hardlinkks and symlinks.
http://fstools.macosforge.org/trac/browser/trunk/src

Liryna:
@andreyloktev , I didn't forget you ! I just have couple of things to do and busy with the pre-release :(
I will try to do it at least at the end of the week.

Maxhy:
@andreyloktev thanks you for this PR. But I'm wondering if this shouldn't be only handled kernel side so user mode file system don't have to care about it. What's the value to handle it partially on user land?
Also it looks to me that you committed some duplicate code, see https://github.com/andreyloktev/dokany/blob/master/dokan_mirror/mirror.c#L526 and following lines which looks duplicated to the previous ones.

andreyloktev:
Hello @Maxhy, sorry for delay. Thanks for your review. Yes, these lines are duplicated. I'll fix it.
Yes, all this code would be able handled on the kernel side if the dokan contained a file size information. But now all file system information is on the user mode side.

andreyloktev:
Hi, everyone!
In r7647817 duplicated code in the MirrorWriteFile was removed.

Liryna:
merged with https://github.com/dokan-dev/dokany/commit/1d53d9ae44e7b95563739729959f764812870d3d

PR #68, opened 2015-09-28
Update install.nsi

justanotheranonymoususer:
Added `/REBOOTOK` for removing the driver. Not tested. Also: should the flag be added to the dlls? Can they be locked too?


Liryna:
Thank you and well found :) !
I never seen dokan.dll lock for now so would not need for now.

PR #70, opened 2015-09-28
Update dokan.c

justanotheranonymoususer:
Typo

### DIFF
--- a/dokan/dokan.c
+++ b/dokan/dokan.c
@@ -573,7 +573,7 @@ DispatchUnmount(
 
 	LeaveCriticalSection(&DokanInstance->CriticalSection);
 
-	// do not notice enything to the driver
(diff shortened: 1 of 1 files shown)


Liryna:
Good :D ?

justanotheranonymoususer:
Pardon?

Liryna:
I wanted to know if you have any other typo to fix.

justanotheranonymoususer:
Oh. That's all for now. Just saw them with a VS extension that highlights typos while browsing that file. I guess that there are many more, if I found so many in just one file :), but it's not convenient to fix these through the GitHub editor, and I don't have a Git client at hand to submit a proper pull request.

Liryna:
Oh ok :D !

PR #72, opened 2015-09-28
Read-only device option

justanotheranonymoususer:
I haven't tested this pull request, because I cannot rebuild the driver. If you could check it, or build the driver and let me check, that would be great.

### DIFF
--- a/dokan/dokan.c
+++ b/dokan/dokan.c
@@ -624,6 +624,9 @@ DokanStart(PDOKAN_INSTANCE Instance)
 	if (Instance->DokanOptions->Options & DOKAN_OPTION_REMOVABLE) {
 		eventStart.Flags |= DOKAN_EVENT_REMOVABLE;
 	}
+	if (Instance->DokanOptions->Options & DOKAN_OPTION_READONLY) {
+		eventStart.Flags |= DOKAN_EVENT_READONLY;
+	}
     
     eventStart.IrpTimeout = Instance->DokanOptions->Timeout;
     
--- a/dokan/dokan.h
+++ b/dokan/dokan.h
@@ -46,6 +46,7 @@ extern "C" {
 #define DOKAN_OPTION_ALT_STREAM	4 // use alternate stream
 #define DOKAN_OPTION_NETWORK	16 // use network drive, you need to install Dokan network provider.
 #define DOKAN_OPTION_REMOVABLE	32 // use removable drive
+#define DOKAN_OPTION_READONLY	64 // use readonly drive
 
 typedef struct _DOKAN_OPTIONS {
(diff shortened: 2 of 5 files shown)


Liryna:
Thanks !
I will test it in the next days :)

Liryna:
Sorry but I wasn't able to make it work :cry: 
`
>mirror.exe /d /s /o /r C:\Users /l m
...
WriteFile : C:\Users\liryna\Desktop\New Text Document.txt, offset 0, length 17
        write 17, offset 0`

Is there something I made wrong ?

justanotheranonymoususer:
Thanks for trying!

> Is there something I made wrong ?

You, as the repo owner, should know better :)
I implemented it based on the `DOKAN_EVENT_REMOVABLE` flag. Don't know whether I did it correctly. We can try a couple of things, e.g. always forcing the `FILE_READ_ONLY_DEVICE` flag and checking whether it has any effect at all.

marinkobabic:
What means you was not able to get it work? 

Step 1: 
Ensure that the device characteristics is set properly using the new parameter. 

Step 2:
Driver should check the device characteristics and respond properly to any CUD operation.

Liryna:
@marinkobabic, I was able to write in a file from the dokan device with the read only option.
I didn't look more after.
I think the Step 2 you describe is not implemented here.

marinkobabic:
@Liryna
You are right. Just put it into the todo list. Should in generally not be a big effort.

justanotheranonymoususer:
I found out that `FILE_READ_ONLY_VOLUME` can be set here:
https://github.com/dokan-dev/dokany/blob/7799f883acd08d75dbcf185d6dd53581f845a3bb/dokan_mirror/mirror.c#L1109-L1113

The solution doesn't seem to be complete, e.g. files cannot be edited, but attributes can still be set, I think (haven't done a through check).

canardos:
As pointed out above, the changes aren't achieving the desired result because the file system is not checking if the device is writable. The flags themselves are not sufficient.

NTFS and FAT etc. use a `IOCTL_DISK_IS_WRITABLE` device req when mounting to determine if the device is writable and set the file system flags accordingly. IO operations are then restricted as appropriate by the file system. The MS fastfat sample follows the same approach.

I've implemented the read-only option in the same style, with `IOCTL_DISK_IS_WRITABLE` and `IOCTL_VOLUME_GET_GPT_ATTRIBUTES` device requests returning read-only and dokan checking write requests at the driver level. It seems to function well and Explorer correctly indicates that the volume is write-protected during write operations.

I'll upload a pull-request tomorrow after I've tested it on Win7 (only tested on Win10 thus far).

Liryna:
I would be glad to review your PR @canardos :) !

Liryna:
I close since https://github.com/dokan-dev/dokany/pull/105 has been merged.

PR #74, opened 2015-09-28
Correctness and performance fixes

justanotheranonymoususer:


justanotheranonymoususer:
Also, a question: who frees the memory here? https://github.com/dokan-dev/dokany/blob/217f3163bdb28ee8928fc02c359939a4dad347e8/dokan_fuse/src/docanfuse.cpp#L462

justanotheranonymoususer:
Here, the check is always false, because the variable is unsigned: https://github.com/dokan-dev/dokany/blob/217f3163bdb28ee8928fc02c359939a4dad347e8/dokan_fuse/src/utils.cpp#L154

Liryna:
Thank you !
DId you faced an issue with MirrorFindFiles ?
Even if the changes are simple in dokan_fuse, did you test it ? (I have no projet with fuse :( )

About https://github.com/dokan-dev/dokany/blob/217f3163bdb28ee8928fc02c359939a4dad347e8/dokan_fuse/src/docanfuse.cpp#L462
DokanMain does not free it, so there is probably a leak here :dancer: 

As you can see dokan_fuse has no maintainer :( the projet was only imported here because abandonned.
So if someone is ready to maintain it :+1: !!!

justanotheranonymoususer:
> Even if the changes are simple in dokan_fuse, did you test it ? (I have no projet with fuse :( )

No. I found the issues with a static analyzer, and they are quite straightforward. I don't believe something can go wrong.

Liryna:
After review, it seems to be safe changes even without test.
Thanks as usual :smiley:

PR #75, opened 2015-09-28
Typo

justanotheranonymoususer:


Liryna:
Thank you, what is the name of your vs extension ?

justanotheranonymoususer:
I use Visual Assist X, which is not free, but Google says that there are many alternatives (Google for "Visual Studio Spell Checker")

Issue #57, opened 2015-09-15
DokanUnmount() always fails


Corillian:
I have not been able to find a way to call DokanUnmount() such that it actually succeeds. Not only that but `dokanctl.exe /u M` also fails. The only way I can unmount the drive is by executing `dokanctl.exe /u M /f` however DokanMain() still doesn't return and additional calls to DokanUnmount() also fail.

I am running a console application and DokanMain() is being called on its own thread with `DOKAN_OPTION_KEEP_ALIVE | DOKAN_OPTION_DEBUG`.

Liryna:
Can you provide some logs ?
https://github.com/dokan-dev/dokany/wiki/How-to-Debug-Dokan

Corillian:
Here's the log from the Visual Studio debug console:

```
The thread 0x15bc has exited with code 0 (0x0).
SftpManager::Initialize() WSAStartup() succeeded
SftpManager::Initialize() libssh2 initialized
Dokan: debug mode on
device opened
mounted: M:\ -> \Volume{d6cc17c5-173a-4085-bce7-964f1e9f5de9}
###Create 0000
SftpOpenDirectory()
CreateFile status = 0
###QueryVolumeInfo 0000
SftpGetVolumeInformation()
###QueryVolumeInfo 0000
SftpGetVolumeInformation()
###Cleanup 0000
SftpCleanup()
###Close 0000
SftpCloseFile()
###Create 0001
   CreateDisposition 0x00000001
CreateFile status = -2
###Create 0002
   CreateDisposition 0x00000001
CreateFile status = -2
###Create 0003
SftpOpenDirectory()
CreateFile status = 0
###QueryVolumeInfo 0003
SftpGetVolumeInformation()
###QueryVolumeInfo 0003
SftpGetVolumeInformation()
###Create 0004
###Cleanup 0003
SftpCleanup()
###Create 0005
SftpOpenDirectory()
###Close 0003
SftpCloseFile()
CreateFile status = 0
SftpOpenDirectory()
CreateFile status = 0
###QueryVolumeInfo 0005
###QueryVolumeInfo 0004
SftpGetVolumeInformation()
SftpGetVolumeInformation()
###QueryVolumeInfo 0004
SftpGetVolumeInformation()
###QueryVolumeInfo 0005
SftpGetVolumeInformation()
###Create 0006
   CreateDisposition 0x00000001
CreateFile status = 0
###Cleanup 0004
###GetFileInfo 0006
[...]

PR #79, opened 2015-09-29
Fixed an issue with DokanUnmount() failing to unmount a drive

Corillian:
This fix also applies to dokanctl.exe /u
Enabled debug information for dokan.dll and dokannp.dll when compiling
in Debug so that I could debug
Issue #57


Liryna:
Thanks for this PR :heart: and taking the time in your issue :D !

To finish to normalizing mount paths, don't you think we should also change the mount function to force the `M:\` ?
https://github.com/Corillian/dokany/blob/InvestigateHang/dokan_mount/mount.c#L263

And also fix the mirror :smile: https://github.com/Corillian/dokany/blob/InvestigateHang/dokan_mirror/mirror.c#L51

Corillian:
Hey sorry got sidetracked with working on other stuff. Yeah I can add normalization to mirror and mount.

Corillian:
Ok I added the normalization to mounter.exe as that seems to be the only place where it matters. Drive letter names are normalized to the format "C:\" upon insertion into `g_MountList` as well as when `g_MountList` is being searched. I also updated this branch to the latest on master. Let me know if I missed anything.

Liryna:
Thank you @Corillian and for taking the time !
I will try to take a look this weekend :)

Liryna:
Perfectly working on my side after test !
Thank a lot @Corillian for your contributions in the dokan project !

Normalizing the mount point will help a lot the dev & users of dokan !

PR #84, opened 2015-10-10
Use _malloca() for debug strings to avoid buffer overflows

nmlgc:
While the sprintf_s functions are supposed to protect from buffer overflows and other errors, they still raise an exception if the buffer is too small, so any debug string longer than 512 characters still crashes the program. In practice, this limit is exceeded very easily, for example, when [matching a file name with a length of MAX_PATH against itself](https://github.com/dokan-dev/dokany/blob/master/dokan/directory.c#L382).

Instead of just increasing the buffer size and/or using snprintf, I prefer to use a combination of _vscwprintf() and _malloca()/_freea(), which computes the final length and then calls either alloca() for strings below 1024 bytes or malloc() for longer ones. This makes sure the string always gets shown in full (well, except for the unlikely case of a malloc() failure) and should eliminate this bug class without impacting performance too much.

### DIFF
--- a/dokan/dokanc.h
+++ b/dokan/dokanc.h
@@ -23,6 +23,7 @@ with this program. If not, see <http://www.gnu.org/licenses/>.
 #define _DOKANC_H_
 
 #include "dokan.h"
+#include <malloc.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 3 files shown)


Liryna:
Thank you !
This is perfectly working !
Always glad to see such contribution and special when it is about helping dokan stability and safety !

PR #88, opened 2015-10-13
use struct stat from cygwin as struct FUSE_STAT for better compatibility

x1596357:
The original struct stat lacks members like st_blksize etc and uses atime/mtime/ctime, which can only have accuracy in seconds. What's more, the difference with the struct FUSE_STAT in DokanFUSE and struct stat in user program causes stack overflows. The new struct FUSE_STAT is ported from msys2 64bits, which is actually ported from cygwin 64bits. With this modification, users with msys/cygwin(64-bit)toolchains can compile their FUSE programs with DokanFUSE library directly without any more modification other than changing their header files. However, for other toolchains like MSVC,mingw and cygwin 32bits, you need to explicitly replace struct stat in your FUSE program to use struct FUSE_STAT from fuse_win.h. Because the implementation of stat.h varies from operating systems and toolchains, it's hard to find a simply way to be compatible with all of them.

### DIFF
--- a/dokan_fuse/include/fuse_win.h
+++ b/dokan_fuse/include/fuse_win.h
@@ -28,13 +28,13 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
-int ntstatus_error_to_errno(int win_res);
-int errno_to_ntstatus_error(int err);
+	int ntstatus_error_to_errno(int win_res);
+	int errno_to_ntstatus_error(int err);
 
-//This stuff is useful only on Windows in MSVC
+	//This stuff is useful only on Windows in MSVC
 #ifdef _MSC_VER
-char** convert_args(int argc, wchar_t* argv[]);
-void free_converted_args(int argc, char **argv);
+	char** convert_args(int argc, wchar_t* argv[]);
+	void free_converted_args(int argc, char **argv);
 #endif
 
 #ifdef __cplusplus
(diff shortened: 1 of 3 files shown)


Liryna:
Hi @x1596357 ,

Thanks alot for the contribution !
Everything seems to be alright :+1:

I just wanted to ask of you could change the place of st_blksize and st_blocks for having the same memory struct than the linux struct stat ?

By the way, I will change dokan fuse to become a dynamic library as you asked.

x1596357:
Hi @Liryna,  
Thanks for your efforts for changing to dokan fuse library.   
Yes, I could make struct FUSE_STAT the same as Linux struct stat but that would require all user programs to explicitly change their struct stat to struct FUSE_STAT, regardless of what compiler they are using. By using a cygwin struct stat, we can at least have msys2 and cygwin being compatible. And since mingw and MSVC have their incomplete struct stat, users who uses them would have to use struct FUSE_STAT anyway. I think the point is the struct stat in dokan fuse needs to match the one from user program.  
What do you think? Should I make it the same as Linux struct stat and require all users to change their source codes or at least make one compiler compatible?

Liryna:
Oh ok ! Well yes, as you made it seems to be the best choice :)
Could you add this build directive to the dokan fuse wiki https://github.com/dokan-dev/dokany/wiki/FUSE ?

Dokan fuse has been changed to dynamic library: https://github.com/dokan-dev/dokany/commit/4953160a92fc7eba1a51ceab3012278ab6768eab
Since I do not use it, could you tell me if it is enough or do we have to add some export ?

x1596357:
Wiki is already updated. 
It's enough. Exporting is not needed.

Liryna:
Perfect ! Thanks a lot ! :heart:

PR #90, opened 2015-10-14
Winfuse readonly

c-klapaucius:
I have added two commits.

The first commit fixes the O_RDONLY parameter, passed to the fuse_win callbacks.
I could apply the same modification to required_share() but I cannot test it at the moment.

The second commit adds partial read-only export to the system. We definitely want to export files with mode 0444, 0111, 0544... as a read-only files.

Both commits were tested with the the "hello" FUSE sample file.


Liryna:
Thank you @c-klapaucius for you contribution !
If you seen any changes that should be done in dokan fuse, I would be happy to review it !

Do you have other example as "hello" FUSE that could help to know the stat of dokan fuse ?

kriswebdev:
Thanks you. I just came to the conclusion that these incorrect flags were partially causing read-only file open issues in encfs4win, along with a coding mistake in encfs4win.
This bug is one cause of https://github.com/freddy77/encfs4win/issues/5. However, encfs4win doesn't support Dokany 0.8.

PR #96, opened 2015-10-26
Add CreateFileEx to DOKAN_OPERATIONS

billziss-gh:
This PR adds a new operation CreateFileEx() to DOKAN_OPERATIONS. The operation can be used to create/open files/directories. The PR does not remove CreateFile/CreateDirectory/OpenDirectory as to not break existing file systems.

The PR introduces the following operation:

    NTSTATUS (DOKAN_CALLBACK *CreateFileEx) (
        LPCWSTR,                // FileName
        DWORD,                  // DesiredAccess
        DWORD,                  // ShareAccess
        DWORD,                  // CreateDisposition
        DWORD,                  // CreateOptions
        DWORD,                  // FileAttributes
        PSECURITY_DESCRIPTOR,   // Reserved
        PDOKAN_FILE_INFO);

This operation is different from the existing CreateFile as follows:
- CreateDisposition uses the constants from NtCreateFile:
 - FILE_SUPERSEDE, FILE_CREATE, FILE_OPEN, FILE_OPEN_IF, FILE_OVERWRITE, FILE_OVERWRITE_IF
- CreateOptions uses the constants from NtCreateFile:
 - FILE_DIRECTORY_FILE, FILE_NON_DIRECTORY_FILE, etc.
- FileAttributes contains only a (new) file's attributes. It is not FlagsAndAttributes as in Win32 CreateFile.
- It has a Reserved parameter that can be used in the future to pass a security descriptor to attach to a (new) file.

The main advantage of using CreateFileEx is that is passes more information to the file system so that it can make correct decisions with respect to flags such as FILE_DIRECTORY_FILE, FILE_NON_DIRECTORY_FILE. It also eliminates the split between CreateFile/CreateDirectory/OpenDirectory.


Liryna:
Nice to see you here Bill :)
I have not found the time to test but at first look:

It seems that your implementations jump all NtStatus checker:
https://github.com/billziss-gh/dokany/blob/CreateFileEx/dokan/create.c#L267

I do not really like to have two fonction in one. Is it possible to refractor the code or can you move this part of the code in a DispatchCreateEx ?
https://github.com/billziss-gh/dokany/blob/CreateFileEx/dokan/create.c#L81
Make the condition here:
https://github.com/billziss-gh/dokany/blob/CreateFileEx/dokan/dokan.c#L333

And add a space here https://github.com/dokan-dev/dokany/pull/96/files#diff-784a504eea06dc6c849c0ca118117afdR109 :smile: 

Otherwise, I have to look more precisly about the utility of the security context.

billziss-gh:
Fair enough on your comment about having two functions in one. I will split them in two.

But I find that placing the condition in DokanLoop is not the appropriate place. The switch is nice and easy to understand there, we should not overcomplicate it with such conditions IMO.

Instead I will place the condition right in the beginning of DispatchCreate. I will make things such that you can easily move the condition to DokanLoop, if you have a strong preference to do it that way.

Liryna:
@billziss-gh Directly at the beginning of dispatchcreate is alright for me too :)

billziss-gh:
Ok, done and tested with winfstest. I believe I incorporated all your proposed changes.

I also made eventInfo a stack variable instead of a malloc allocation, which seems unnecessary to me (only in DispatchCreateEx though, I did not touch DispatchCreate other than the conditional at the beginning).

Some comments regarding the security descriptor. The Win32 CreateFile allows one to pass a security descriptor to attach to a newly created file. Here is the Win32 CreateFile prototype:

    HANDLE WINAPI CreateFile(
      _In_     LPCTSTR               lpFileName,
      _In_     DWORD                 dwDesiredAccess,
      _In_     DWORD                 dwShareMode,
      _In_opt_ LPSECURITY_ATTRIBUTES lpSecurityAttributes,
      _In_     DWORD                 dwCreationDisposition,
      _In_     DWORD                 dwFlagsAndAttributes,
      _In_opt_ HANDLE                hTemplateFile
    );

The documentation has this to say about lpSecurityAttributes:

> The lpSecurityDescriptor member of the structure specifies a SECURITY_DESCRIPTOR for a file or device. If this member is NULL, the file or device associated with the returned handle is assigned a default security descriptor.
> CreateFile ignores the lpSecurityDescriptor member when opening an existing file or device.

So when creating a new file it is possible to attach a security descriptor to it. This capability does not seem to exist in Dokan currently. This is why I added the Reserved security descriptor parameter, so that we can pass it at a later time from the driver.

Liryna:
I close since #104 has been merged.

PR #105, opened 2015-11-11
Add write protect device option

canardos:
These changes add a write-protect option to the dokan framework and mirror sample. 

The write-protect functionality is implemented at the driver level (i.e. kernel space) and doesn't require any changes to file system implementations beyond specifying the write-protect option during mounting.

Requests that would result in write operations are intercepted by the driver, avoiding any unnecessary calls into user space. One exception to this is `FILE_OPEN_IF` requests. At the driver level, we don't know if a file exists without calling into the file system implementation. We therefore don't know whether to fail `FILE_OPEN_IF` requests.

The current solution to this is to substitute `FILE_OPEN` for `FILE_OPEN_IF` before passing the request to the user mode code, and if we receive a status not found, convert to a status write protected.

The basic functionality has been tested from explorer on Win7 x64 and Win10 x64, but certain paths are beyond my current test setup (e.g. `IOCTL_DISK_IS_WRITABLE` and `IOCTL_VOLUME_GET_GPT_ATTRIBUTES` calls, `FILE_OPEN_IF`).

Please take a look.


Liryna:
Tested and work perfectly for me :trophy: !
Thank a lot for your contribution @canardos , I think you made implementation how it has to be done.

PR #107, opened 2015-11-12
Remove potential memory leak in DokanCreateDiskDevice on allocation failure.

canardos:
The existing code assumes that if one of `SymbolicLinkName` `DiskDeviceName` or `FileSystemDeviceName` fails allocation, then the subsequent allocations also failed and does not free them. This may not be the case, resulting in a memory leak.


canardos:
I've noticed other execution paths in this function that also fail to free the memory. Closing and will resubmit with further changes

canardos:
Can't reopen due to [361](https://github.com/isaacs/github/issues/361), resubmitted.

PR #108, opened 2015-11-12
Remove potential memory leak in DokanCreateDiskDevice on allocation failure (resubmit)

canardos:
The existing code assumes that if one of `SymbolicLinkName` `DiskDeviceName` or `FileSystemDeviceName` fails allocation, then the subsequent allocations also failed and does not free them. This may not be the case, resulting in a memory leak.

Other failure paths also fail to free these items.

Note: Resubmission of [pull 107](https://github.com/dokan-dev/dokany/pull/107).


Liryna:
That is well seen :) ! even code analyser haven't found  it!

PR #114, opened 2015-11-29
get_disk_free_space() should not return -EINVAL - this confuses Windo…

c-klapaucius:
…ws Explorer

and results in an error message (Incorrect MS-DOS function)


Liryna:
Thank you ! :+1:

PR #127, opened 2015-12-15
Thread win32 error code now gets reset before the ZwCreateFile callback is called

Corillian:
See https://github.com/dokan-dev/dokany/commit/1039c0adcd413733399995df0169010d17c01dde for more information


Liryna:
:+1: Thank you @Corillian !

PR #130, opened 2016-01-04
Fix the Dokan FUSE library

toksaitov:
Hi,

I have found a number of problems with the FUSE wrapper while trying to port my FUSE file system from *nix to Windows. Here's my attempt to fix them. Please, check commit messages for more information.

Sample use cases before
http://i.imgur.com/kDFPyPW.gifv
…and after
http://i.imgur.com/dcUGsAd.gifv

Best regards,
Dmitrii Toksaitov

### DIFF
--- a/dokan_fuse/include/fusemain.h
+++ b/dokan_fuse/include/fusemain.h
@@ -95,6 +95,10 @@ public:
 	int do_create_file(LPCWSTR FileName, DWORD Disposition, DWORD share_mode, DWORD Flags,
 		PDOKAN_FILE_INFO DokanFileInfo);
 
+    int do_delete_directory(LPCWSTR file_name, PDOKAN_FILE_INFO dokan_file_info);
+
+    int do_delete_file(LPCWSTR file_name, PDOKAN_FILE_INFO dokan_file_info);
+
 	int convert_flags(DWORD Flags);
 
 	int resolve_symlink(const std::string &name, std::string *res);
--- a/dokan_fuse/src/docanfuse.cpp
+++ b/dokan_fuse/src/docanfuse.cpp
@@ -197,8 +197,8 @@ FuseCreateFile(LPCWSTR FileName, PDOKAN_IO_SECURITY_CONTEXT SecurityContext,
     }
   }
 
-  return -win_error(impl->create_file(FileName, CreateDisposition, ShareAccess,
-                                      DesiredAccess, FileAttributes,
+  return -win_error(impl->create_file(FileName, DesiredAccess, ShareAccess,
+                                      CreateDisposition, FileAttributes,
                                       DokanFileInfo));
 }
 
--- a/dokan_fuse/src/docanfuse.cpp
+++ b/dokan_fuse/src/docanfuse.cpp
@@ -197,9 +197,9 @@ FuseCreateFile(LPCWSTR FileName, PDOKAN_IO_SECURITY_CONTEXT SecurityContext,
     }
   }
 
-  return -win_error(impl->create_file(FileName, DesiredAccess, ShareAccess,
-                                      CreateDisposition, FileAttributes,
-                                      DokanFileInfo));
+  return impl->create_file(FileName, DesiredAccess, ShareAccess,
+                           CreateDisposition, FileAttributes,
+                           DokanFileInfo);
 }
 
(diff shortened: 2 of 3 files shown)


Liryna:
Hello @toksaitov ,

Always great to see people using the FUSE API !

About https://github.com/toksaitov/dokany/commit/030fcf791e8e830ce1b98072f6428da7c4fcaa2a, it is great that you have add it ! 
```
According to Windows development documentation for the
`CreateFile` function, for a similar case with a disposition set to
`FILE_CREATE`, we have to set the last error to `ERROR_FILE_EXISTS`.
```
This is currently not handle :smiley: but I will add it.
```
According to Windows development documentation for the `CreateFile` function, in
case a call to `CreateFile` with a `CreateDisposition` set to `FILE_OPEN` or
`FILE_OVERWRITE` can’t find a file to open or truncate, we have to set the last
error to `ERROR_FILE_NOT_FOUND `.
```
We are always setting `FILE_DOES_NOT_EXIST` in case of createfile error. https://github.com/dokan-dev/dokany/blob/master/dokan/create.c#L219
I will look if this have to be changed.

For https://github.com/toksaitov/dokany/commit/3f2f5de91554802c3bf4c27c5c01cc7e76b45e7d, could you show me the ReactOS code ? I have no fuse sample so I will trust you on this.

https://github.com/toksaitov/dokany/commit/566ff428631162d25322a32c82ef88290b09a0fe#diff-a8f32b6fd8169cac1606e8d8b0f497c1R400 Does calling delete during createfile is the correct workflow for FUSE ?

Very great PR ! Otherwise than dokan_fuse issue, did you had to change a lot of your fuse driver to make it work with dokan ?

toksaitov:
Hi, @Liryna 

>This is currently not handle… We are always setting FILE_DOES_NOT_EXIST in case…

I have taken a look at Dokan's internals. I am sorry, my mistake. I wasn't aware the last error is not passed as it is through the chain.

It's not an expensive operation, though. Maybe, it can be used in the future or could serve as self documentation in the current FUSE wrapper. The last word is yours.

>…could you show me the ReactOS code ? I have no fuse sample so I will trust you on this.

The fix was based on this [code](http://doxygen.reactos.org/d3/d60/ntoskrnl_2io_2iomgr_2file_8c_a4ab64d565c29a961b5b9001aca6927cb.html).

>Does calling delete during createfile is the correct workflow for FUSE ?

I am calling delete functions from the `Cleanup` handler. According to [this](https://github.com/dokan-dev/dokany/wiki/FUSE), `Cleanup` is similar to FUSE `release` and `releasedir` callbacks, and according to [this](https://github.com/libfuse/libfuse/blob/master/include/fuse.h#L257) it seems to be OK to call delete callbacks after closing a file (which is done by `Cleanup`).

>…did you had to change a lot of your fuse driver to make it work with dokan ?

The only change was to add an `fsync` callback to always return a success code even though my system doesn't use it. It seems that *nix FUSE implementations were doing something similar for undefined `fsync` handlers.

Maxhy:
Just talking about FUSE in general here (but your PR looks great, I will let Liryna handle the review :smile:).
I don't have enough experience about fuse and I also believe a fuse mirror sample would be great, with associated documentation on the wiki of course.
Do you think you could handle this @toksaitov? This would consolidate fuse support and ideally allow fuse unit testing.

Liryna:
Thank @toksaitov for your answer, it is more clear!

I will look later for making dokan handle the new setlasterror that you have added.

I am ready to accept your PR as it is. But as requested by @Maxhy, it would be very great if you could add a fuse mirror sample :heart: to your PR.

toksaitov:
@Liryna @Maxhy
>…a fuse mirror sample would be great

Agreed. I will give it a try, but a bit later and in a separate PR.

Thank you.

Liryna:
That's Good news @toksaitov ! Can't wait to see it :) !

Liryna:
Just for information @toksaitov , I checked dokan code about SetLastError. We only use the value for OPEN_ALWAYS & CREATE_ALWAYS. Otherwise the value is even not used BUT the kernel is correctly notified for the different case as described in the createfile documentation.

So `SetLastError(ERROR_FILE_EXISTS);` and `SetLastError(ERROR_FILE_NOT_FOUND);` are not used.

PR #139, opened 2016-01-14
VPB read only members should not be changed

BoRiSGz:
The RealDevice and Flags members are read only members of the VPB structure (https://msdn.microsoft.com/en-us/library/windows/hardware/ff625892%28v=vs.85%29.aspx).
VPB_MOUNTED is set by the I/O manager immediately after IRP_MN_MOUNT_VOLUME is completed with a successful status.
RealDevice should point to the disk device already as this point.

### DIFF
--- a/sys/fscontrol.c
+++ b/sys/fscontrol.c
@@ -325,11 +325,8 @@ VOID DokanInitVpb(__in PVPB Vpb, __in PDEVICE_OBJECT DiskDevice,
                   __in PDEVICE_OBJECT VolumeDevice) {
   if (Vpb != NULL) {
     Vpb->DeviceObject = VolumeDevice;
-    Vpb->RealDevice = DiskDevice;
-    Vpb->Flags |= VPB_MOUNTED;
     Vpb->VolumeLabelLength = (USHORT)wcslen(VOLUME_LABEL) * sizeof(WCHAR);
-    RtlStringCchCopyW(Vpb->VolumeLabel,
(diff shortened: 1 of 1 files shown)


Liryna:
Thank you @BoRiSGz you are right this should not be change by us !

About your changes, I guess the second parameter of `RtlStringCbCopyW` should be `Vpb->VolumeLabelLength` and param DiskDevice can be remove.

However your pull request breaks dokan so I cannot accept it as it is. It make the device mount->unmont instantly.

@Maxhy has already try this changes and moved DokanInitVpb in IRP_MN_MOUNT_VOLUME but face the same issue of instant unmount.
If you have an idea @BoRiSGz why this is happening, it would be very helpful !

BoRiSGz:
@liryna
You are correct, I meant to pass Vpb->VolumeLabelLength as the second parameter.
I'll push the changes when I have the chance.

@Maxhy can you please elaborate on this?
What do you mean instant unmount?
Dies the system send an instant dismount control to the device?

Maxhy:
Thank you for your PR @BoRiSGz, it has the benefits to point out the core issue.
I will close your PR as it requires a more advanced approach (to answer your question, it is not the system that unmount the device but dokan dll or other threads), but be sure it wasn't done for nothing! See the current discussion it generated on @marinkobabic PR #140 which re-embed your fix as you can see. Your feedback would be helpful.
Thanks again.

PR #140, opened 2016-01-14
Fix mounting for socket connection

marinkobabic:
We need to think what the best way is to wait until the drive is mounted. My approach works but it looks like a hack. Maybe Maxhy has an better idea.

### DIFF
--- a/dokan/dokan.c
+++ b/dokan/dokan.c
@@ -249,6 +249,12 @@ int DOKANAPI DokanMain(PDOKAN_OPTIONS DokanOptions,
     return DOKAN_MOUNT_ERROR;
   }
 
+  
+  // wait until the device is realy mounted
+  if (DokanOptions->Options & DOKAN_OPTION_MOUNT_MANAGER) {
+	  WaitForMount(instance);
+  }
+
   // Here we should have been mounter by mountmanager thanks to
   // IOCTL_MOUNTDEV_QUERY_SUGGESTED_LINK_NAME
   DbgPrintW(L"mounted: %s -> %s\n", instance->MountPoint, instance->DeviceName);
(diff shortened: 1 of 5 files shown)


dokan/dokan.c
```diff
@@ -249,6 +249,12 @@ int DOKANAPI DokanMain(PDOKAN_OPTIONS DokanOptions,
     return DOKAN_MOUNT_ERROR;
   }
 
+  
+  // wait until the device is realy mounted
+  if (DokanOptions->Options & DOKAN_OPTION_MOUNT_MANAGER) {
```
Maxhy: The issue here is not related to Mount manager but to I/O manager. It is general to all Dokan mount method so this check shouldn't be done.

dokan/dokan.c
```diff
@@ -249,6 +249,12 @@ int DOKANAPI DokanMain(PDOKAN_OPTIONS DokanOptions,
...
+	  WaitForMount(instance);
```
Maxhy: It is an interesting way to solve it :).
But as you said it looks like a hack. Wouldn't be better to handle this wait through [RegisterDeviceNotification](https://msdn.microsoft.com/fr-fr/library/windows/desktop/aa363431%28v=vs.85%29.aspx)  to be notified of FSRTL_VOLUME_MOUNT?

sys/event.c
```diff
@@ -127,11 +127,13 @@ RegisterPendingIrpMain(__in PDEVICE_OBJECT DeviceObject, __in PIRP Irp,
   irpEntry->Flags = Flags;
 
   // Update the irp timeout for the entry
+  ExAcquireResourceExclusiveLite(&vcb->Dcb->Resource, TRUE);
```
Maxhy: This and associated release call will fail if vcb is null (see vcb check next line).

sys/fscontrol.c
```diff
@@ -380,7 +378,26 @@ DokanDispatchFileSystemControl(__in PDEVICE_OBJECT DeviceObject,
...
+	  try {
+		  VolumeFileObject = IoCreateStreamFileObjectLite(NULL, vcb->DeviceObject);
+	  } except(EXCEPTION_EXECUTE_HANDLER) {
+		  status = GetExceptionCode();
+		  DDbgPrint("   IoCreateStreamFileObjectLite failed with status code %d\n", status);
+	  }
+
+	  if (status == STATUS_SUCCESS) {
+		  // notify all application about volume mount
+		  FsRtlNotifyVolumeEvent(VolumeFileObject, FSRTL_VOLUME_MOUNT);
```
Maxhy: Wow, I thought this was automatically done by I/O manager once we returned success from IRP_MN_MOUNT_VOLUME. Looks like it is needed indeed (see https://github.com/dokan-dev/dokany/pull/140/files#r49822003).

dokan/dokan.c
marinkobabic: Here we should maybe already start to separate the mounting from device creation. Will keep thinking how it can be solved.

marinkobabic:
DokanLoop should remain like it was

 if (lastError == ERROR_NO_SYSTEM_RESOURCES)

dokan/dokan.c
```diff
@@ -369,7 +437,7 @@ UINT WINAPI DokanLoop(PDOKAN_INSTANCE DokanInstance) {
     if (!status) {
       lastError = GetLastError();
       DbgPrint("Ioctl failed for wait with code %d.\n", lastError);
-      if (lastError == ERROR_NO_SYSTEM_RESOURCES) {
+      if (lastError == ERROR_DEV_NOT_EXIST) {
```
marinkobabic: This if (lastError == ERROR_NO_SYSTEM_RESOURCES)

Should remain. ERROR_DEV_NOT_EXIST is wrong here

sys/fscontrol.c
marinkobabic: Documentation says the we must RegisterDeviceNotification to be able to notify the applications using the code above.

dokan/dokan.c
marinkobabic: RegisterDeviceNotification needs a window handle. It can be realized using a hidden window but it seems again to me to be a hack.

joepperkins: The getdesktopwindow should work for your handle.

https://msdn.microsoft.com/en-us/library/windows/desktop/ms633504(v=vs.85).aspx


Sent on a Sprint Samsung Galaxy Note® 3


-------- Original message --------

marinkobabic: An handle is needed which will be receiver of the messages. If Desktop handle is passed I have no option to access the received messages.


Von Samsung Mobile gesendet


-------- Ursprüngliche Nachricht --------
Von: joepperkins
Datum:15.01.2016 17:49 (GMT+01:00)
An: dokan-dev/dokany
Cc: Babic Marinko
Betreff: Re: [dokany] Fix mounting for socket connection (#140)


In dokan/dokan.c<https://github.com/dokan-dev/dokany/pull/140#discussion_r49875768>:

> @@ -249,6 +249,12 @@ int DOKANAPI DokanMain(PDOKAN_OPTIONS DokanOptions,
>      return DOKAN_MOUNT_ERROR;
>    }
>
> +
> +  // wait until the device is realy mounted
> +  if (DokanOptions->Options & DOKAN_OPTION_MOUNT_MANAGER) {
> +       WaitForMount(instance);


The getdesktopwindow should work for your handle. https://msdn.microsoft.com/en-us/library/windows/desktop/ms633504(v=vs.85).aspx Sent on a Sprint Samsung Galaxy Note® 3
-------- Original message -------- From: Marinko <notifications@github.com> Date: 01/15/2016 1:15 AM (GMT-08:00) To: dokan-dev/dokany <dokany@noreply.github.com> Subject: Re: [dokany] Fix mounting for socket connection (#140<https://github.com/dokan-dev/dokany/pull/140>) In dokan/dokan.c<#140 (comment)<https://github.com/dokan-dev/dokany/pull/140#discussion_r49833461>>:
@@ -249,6 +249,12 @@ int DOKANAPI DokanMain(PDOKAN_OPTIONS DokanOptions, return DOKAN_MOUNT_ERROR; } + + // wait until the device is realy mounted + if (DokanOptions->Options & DOKAN_OPTION_MOUNT_MANAGER) { + WaitForMount(instance);
RegisterDeviceNotification needs a window handle. It can be realized using a hidden window but it seems again to me to be a hack. - Reply to this email directly or view it on GitHub<https://github.com/dokan-dev/dokany/pull/140/files#r49833461>.

-

PR #141, opened 2016-01-15
Fix mounting to get socketConnection Work

marinkobabic:
Fix mounting to get socketConnection Work. Sorry had some issue with my fork. Therefore the new request :smile:

### DIFF
--- a/sys/notification.c
+++ b/sys/notification.c
@@ -250,7 +250,7 @@ VOID NotificationLoop(__in PIRP_LIST PendingIrp, __in PIRP_LIST NotifyEvent) {
     }
 
     if (IoSetCancelRoutine(irp, NULL) == NULL) {
-	  DDbgPrint("IoSetCancelRoutine return NULL\n");
+      DDbgPrint("IoSetCancelRoutine return NULL\n");
       // Cancel routine will run as soon as we release the lock
       InitializeListHead(&irpEntry->ListEntry);
       irpEntry->CancelRoutineFreeMemory = TRUE;
--- a/sys/timeout.c
+++ b/sys/timeout.c
@@ -327,19 +327,19 @@ Routine Description:
 
 --*/
 {
-	DDbgPrint("==> DokanStopCheckThread");
+  DDbgPrint("==> DokanStopCheckThread");
 
-	if (KeSetEvent(&Dcb->KillEvent, 0, FALSE) > 0 && Dcb->TimeoutThread) {
-		DDbgPrint("Waiting for Timeout thread to terminate.\n");
-		ASSERT(KeGetCurrentIrql() <= APC_LEVEL);
-		KeWaitForSingleObject(Dcb->TimeoutThread, Executive, KernelMode, FALSE,
-			NULL);
-		DDbgPrint("Timeout thread successfully terminated.\n");
-		ObDereferenceObject(Dcb->TimeoutThread);
-		Dcb->TimeoutThread = NULL;
-	}
+  if (KeSetEvent(&Dcb->KillEvent, 0, FALSE) > 0 && Dcb->TimeoutThread) {
+    DDbgPrint("Waiting for Timeout thread to terminate.\n");
+    ASSERT(KeGetCurrentIrql() <= APC_LEVEL);
+    KeWaitForSingleObject(Dcb->TimeoutThread, Executive, KernelMode, FALSE,
+                          NULL);
+    DDbgPrint("Timeout thread successfully terminated.\n");
+    ObDereferenceObject(Dcb->TimeoutThread);
+    Dcb->TimeoutThread = NULL;
+  }
 
-	DDbgPrint("<== DokanStopCheckThread");
+  DDbgPrint("<== DokanStopCheckThread");
 }
 
 NTSTATUS


Liryna:
Hi @marinkobabic ,

Thank you for your contribution ! This will fix a major issue :+1: 

First review:
* Is it possible that `WaitForMount` does not exist because the device did'nt mount for a reason ? 
* What happen when `IoCreateStreamFileObjectLite`` fail ? Application are just notified and system keep stable ?
* Does `VolumeFileObject` should cleaned with `ObDereferenceObject` ?

I will test it during my break otherwise.

marinkobabic:
Hi Liryna

* The mount can fail. In our case we actually return STATUS_SUCCESS when it's our device. So it should in theory never fail. I'm still thinking about a different approach to solve this issue, but this takes time
* FsRtlNotifyVolumeEvent is actually not used and it has no effect. The application will remain stable. My expectation was that the explorer will refresh the drive list. Need to check if a broadcast is really sent or not. Therefore I have added more logging information
* Yes. You are right. It's changed

We have done some progress but the WaitForMount must be refactored so that it considers all the possible situations.

Liryna:
Ok that make it more clear and thanks for the changes.

I face an issue with the mirror:
```
## Attempting to add SE_SECURITY_NAME privilege to process token ##
  success: privilege already present
Dokan: debug mode on
Dokan: use stderr
device opened
WaitForMount Ioctl failed for wait with code 1.
WaitForMount we got different error 0
mounted: m -> \Volume{d6cc17c5-1734-4085-bce7-964f1e9f5de9}
Mounted
Ioctl failed for wait with code 1.
Ioctl failed for wait with code 1.
Ioctl failed for wait with code 1.
Ioctl failed for wait with code 1.
Ioctl failed for wait with code 1.
Thread will be terminated
Thread will be terminated
Thread will be terminated
Thread will be terminated
Thread will be terminated
Unmounted

unload
Success
```

Liryna:
Thank you for the changes and your anwers !

I have an error with the mirror:
```
## Attempting to add SE_SECURITY_NAME privilege to process token ##
  success: privilege already present
Dokan: debug mode on
Dokan: use stderr
device opened
WaitForMount Ioctl failed for wait with code 1.
WaitForMount we got different error 0
mounted: m -> \Volume{d6cc17c5-1734-4085-bce7-964f1e9f5de9}
Mounted
Ioctl failed for wait with code 1.
Ioctl failed for wait with code 1.
Ioctl failed for wait with code 1.
Ioctl failed for wait with code 1.
Ioctl failed for wait with code 1.
Thread will be terminated
Thread will be terminated
Thread will be terminated
Thread will be terminated
Thread will be terminated
Unmounted

unload
Success
```

marinkobabic:
What os are you using?

Please can you provide the log of the driver?



Von Samsung Mobile gesendet


-------- Ursprüngliche Nachricht --------
Von: Liryna
Datum:15.01.2016 12:16 (GMT+01:00)
An: dokan-dev/dokany
Cc: Babic Marinko
Betreff: Re: [dokany] Fix mounting to get socketConnection Work (#141)


Hi @marinkobabic<https://github.com/marinkobabic> ,

Thank you for your contribution ! This will fix a major issue [:+1:]

First review:

  *   Is it possible that WaitForMount does not exist because the device did'nt mount for a reason ?
  *   What happen when IoCreateStreamFileObjectLite` fail ? Application are just notified and system keep stable ?
  *   Does VolumeFileObject should cleaned with ObDereferenceObject ?

I will test it during my break otherwise.

-

Liryna:
I am testing on Win10.
Kernel logs: http://pastebin.com/sr8WPADN

marinkobabic:
Have you set the option to use the mount manager?

Liryna:
:anguished: Oh 
Here is with mount manager option (I have the same behaviour):
```
>mirror.exe /r C:\Users /l m /g /d /s
## Attempting to add SE_SECURITY_NAME privilege to process token ##
  success: privilege already present
Dokan: debug mode on
Dokan: use stderr
device opened
WaitForMount Ioctl failed for wait with code 1.
mounted: m -> \Volume{d6cc17c5-1737-4085-bce7-964f1e9f5de9}
Mounted
Ioctl failed for wait with code 1.
Thread will be terminated
Ioctl failed for wait with code 1.
Thread will be terminated
Ioctl failed for wait with code 1.
Ioctl failed for wait with code 1.
Ioctl failed for wait with code 1.
Thread will be terminated
Thread will be terminated
Thread will be terminated
Unmounted

unload
Success
```

Kernel logs: http://pastebin.com/HgNJG16Y

Maxhy:
For me on Win8.1 x64 it is not automatically unmount but it doesn't seems to fix socket issue at the end (I tried with ping.exe).
Anyway this mount process should be improved as suggest here and by @BoRiSGz.
I was thinking about improving WaitForMount but I can only think about similar solution (sending specific IOCTL to Dokan driver => will be similar to what is currently done with IOCTL_EVENT_INFO, waiting kernel side at DokanStart for IRP_MN_MOUNT_VOLUME completion? Not sure it would be better...).

marinkobabic:
@Maxhy 
I tried the provided code and it woked fine. Also nslookup worked then fine. 

The challenge is to get it work on Windows 10. For some reason IRP_MN_MOUNT_VOLUME is not sent even if the filesystem is registered. It takes the /FileSystem/RAW for mounting instead of dokan. RAW is on the last position of the registered filesystems, so it must take dokan filesystem first. 

After this part is done, we can go on with optimization.

marinkobabic:
@Liryna 
Please do the tests without the option /g (without MountManager) and check if the ping and the socket connection work properly. You can test it on Windows 10 or Windows 8, makes no difference.

@Maxhy 
I think all will work now as expected, but without the MountManager. There are a lot of things which must be considered. http://www.osronline.com/article.cfm?id=107. For some reason after the changes I have done, when MountManager is used, the SymbolikLink is not created. Also the MountManager does not add the Link into the registry MountedDevices. It would be great if you could check this part, but actually it has no high priority if all works also without the MountManager.

Maxhy:
@marinkobabic I took a quick look on your changes. I believe mount manager is not working as excepted because DokanSendVolumeArrivalNotification is still called on DokanCreateDiskDevice whereas it should now be on DokanMountVolume. Just guessing, I cannot give a try before several days unfortunately.

Thank you :).

Liryna:
I made the test without mount manager with your changes @marinkobabic .

The mount is extramly slow but ping seems to work !
In the kernel logs there 12 secondes without logs at line 51/52 even if I refresh explorer without stopping.

Kernel logs: http://pastebin.com/rpP72HR8
Mirror:
```
C:\Users\liryna\Desktop\dokany2\x64\Win10 Debug>mirror.exe /r C:\Users /l m /d /s
## Attempting to add SE_SECURITY_NAME privilege to process token ##
  success: privilege already present
Dokan: debug mode on
Dokan: use stderr
device opened
mounted: m -> \Volume{d6cc17c5-1735-4085-bce7-964f1e9f5de9}
Mounted
###Create 0000
###Create 0000
###QueryVolumeInfo -001
###QueryVolumeInfo -001
###QueryVolumeInfo -001
###Create 0002
CreateFile : C:\Users\
CreateFile : C:\Users\
CreateFile : C:\Users\
  AccountName: liryna, DomainName: DESKTOP-RBPTIQA
  AccountName: liryna, DomainName: DESKTOP-RBPTIQA
        ShareMode = 0x0
        AccessMode = 0x100000
        SYNCHRONIZE
        FlagsAndAttributes = 0x2000000
        FILE_FLAG_BACKUP_SEMANTICS
        ShareMode = 0x7
        FILE_SHARE_READ
        FILE_SHARE_WRITE
        FILE_SHARE_DELETE
        OPEN_EXISTING

        AccessMode = 0x80
        FILE_READ_ATTRIBUTES
  AccountName: liryna, DomainName: DESKTOP-RBPTIQA
...
```

marinkobabic:
What is happening is that we are talking to the volume using IOCTL_EVENT_WAIT so that the driver has threads to process driver requests. The problem is that the operating system want's to get some volume information at this time, because of our request. But there are not threads available to process it (we are the first request to the mounted volume). This way we cause the delay. 

Actually what I have changed is, the requests are processed inside of the driver as long as there are no threads arrived by user mode application to process the incoming requests.

Two issue are still open:

- Explorer does not refresh the drive automatically. This means when we navigate inside of the drive and mirror is closed, then the explorer should be closed automatically.
- The mount manager implementation does still not work

I have not investigated the mount manager issue yet. Maybe the following code helps to understand what is going on behind the scenes

http://doxygen.reactos.org/d9/df4/drivers_2filters_2mountmgr_2device_8c_source.html
https://github.com/hacksysteam/WRK-1.2/tree/master/base/ntos

Just additional information. The FsRtlNotifyVolumeEvent does not work without a PDO (physical device object), actually we have only an FDO (functional device object). Also the IRP_MN_QUERY_DEVICE_RELATIONS must be implemented in a proper way, to get the FsRtlNotifyVolumeEvent work. To create the PDO, DokanRegisterDeviceInterface must be executed and so on..I would suggest to solve the mount manager issue as first and then to go on. I suppose this version will remain beta as long the two issues above are not solved.

Thanks
Marinko

Maxhy:
Working much better now thank you.
About mount manager, I fixed it locally so don't worry about it, this is quickly corrected (following my previous comment instructions).
I don't think IRP_MN_QUERY_DEVICE_RELATIONS is necessary here or I'm missing some information. Do you have some sources link to share to instruct myself?

I'm gonna clean some code and merge it as FsRtlNotifyVolumeEvent issue is not directly related to this PR and already present in master.

marinkobabic:
IRP_MN_QUERY_DEVICE_RELATIONS is not needed. You have moved the arrival notification to DokanMountVolume so this fixed the problem with mount manager. No other changes needed. :-) to get it work. FsRtlNotifyVolumeEvent works the same way on windows. You can debug it.

The creation of drive letter can now be moved as a separate step. DefineDosDevice can notify the explorer. If we have the impersonation flag this command can be used in user mode and Explorer should refresh properly. 

Wondering why mount manager does not refresh the explorer.

Maxhy:
Regarding your last comment, I don't know why Mount Manager doesn't refresh explorer, I thought it was part of its job..
So you want to move back mount point creation to user mode to use DefineDosDevice? I would like to avoid that if possible.

For reminder we can keep NotifyAllApps (dokan driver) and NotifyExplorer (dokan lib) but those should be removed once explorer notification corrected.
Anyway we should merge it but the git history is a mess. Could you move DokanSendVolumeArrivalNotification/DokanCreateMountPoint to DokanMountVolume and then cleanup git history to have an unique commit?

eg:
```
   // Create mount point for the volume
   if (dcb->UseMountManager) {
      status = DokanSendVolumeArrivalNotification(dcb->DiskDeviceName);
      if (!NT_SUCCESS(status)) {
         DDbgPrint("  DokanSendVolumeArrivalNotification failed: 0x%x\n", status);
      }
   }
   DokanCreateMountPoint(dcb);
```

Thank you.

marinkobabic:
Wilk do the change in the evening. DefineDosDevice in usermode is only needed when you want to mount the drive for specific user (not visible to other user). In other case you use the mount manager. This are two separate ways to mount. Not going back to mount service like it was before. 

Do you see another way to do a user specific mount?

Maxhy:
Ok thank you, I'm looking forward for your cleanup changes so it can be merged in a proper way.

About mounting to current user only, this shouldn't impact mount manager decision. Mount manager should remains optional, some people want to mount a volume for all users but to not use mntmgr. Indeed DefineDosDevice is justified here for this purpose.
I will handle this part and do appropriate testing, please don't include such behavior on this PR changes :). Then you can also remove NotifyAllApps / NotifyExplorer code I believe.

Maxhy:
Ok I did the change on kernel mode. Because everything was already here, you will understand by seeing the commit.
This means explorer is still not notified, we could fix it by moving back symbolic link creation to user mode but it is not a real solution as the problem will remains when using mount manager...
But, and this is a big BUT, we cannot mount the drive for a specific user but only for specific session (current session), and not Windows Logon session but authenticated session (identified by Authentication Id). We mixed-up some concepts on previous discussion.
Concretely it means that if you mount a drive using Run As or UAC elevation, it will not be available through explorer of the same logon session. It also means that impersonation is not possible (maybe we could found a way to mount for a specific Authentication Id other than the one in calling thread but I'm not sure this would be useful for anyone..).
See 3083fdda2e04d8fc798a7d6cb9eafceafa8bdbaa.

marinkobabic:
Hi Maxhy

Have synced now the changes so that you can merge.

Related to the discussion of user session, I expected that you will change the code in this way :smile: Run as administrator is a different session, this is clear. And you will not see the drive there. It behaves like mapped network drive and this is fine so far. So we are just missing a way to notify the explorer. 

Other guys have solved this problem in writing an "icon overlay handler" for the explorer which was listening to messages sent by driver. This way it works 100% in every case. The problem is that explorer is limited to 15 icon overlay handler and those are executed in the name order. Dropbox, Svn, OneDrive, Sharepoint etc. have also icon overlay handlers.

I think the DefineDosDevice has the power to create the link and to notify the explorer. For some reason explorer listens to this command and it's  WM_DEVICECHANGE broadcast, but not if you send the broadcast your self.

But let's keep the code like you changed it and go on with the investigations. Plug & Play devices do notify the explorer as well, so we need to do it the same way, somehow.

Thanks
Marinko

Maxhy:
I agree.
Could you just clean up the git history to have only one unique commit (rework history locally then force push on your repo, it is supported by github). This to keep it clean when possible. Thank you.

marinkobabic:
You can close this pull request. I have created a new one. I just realized that I hate the visual studio 2015 git integration

Maxhy:
OK I merged it after fixed some mistakes that wasn't present on original PR. I believe this is because you copy/pasted your changes.
I will take the opportunity to talk about that, when we encounter this situation, the PR shouldn't be closed and a new one created, but git history cleaned using git commands to squash your commits for instance and then changes force pushed to your branch. This avoid splitting the discussion and ensure you will not miss something due to copy/past. Nothing mandatory here, just to let you know in case you are not familiar with git :smile:.

Anyway, thank you again for this change and the constructive discussion, this is a major issue that I'm glad to be fixed.

I just noticed a small issue few minutes after the merge... Mounting a network drive sometime fail due to timing issue on mount irp. That's because the named device is the volume in this case and not the disk. It also probably means that UNC registration should be done at mount and not disk device init.

marinkobabic:
I did interactive rebase and squash but the sync button in visual studio took the previous commits of other contributors and added thise to my pull request. The only tool I trust is the git shell.

In case of network drive you are right. I didn't test it.

Liryna:
Thank you @marinkobabic for your contribution as always ! Very glad to have you with us !

Do you have an idea how to fix the network driver mount ? Or even a patch :smiley: ?

PR #153, opened 2016-02-01
Call FindFiles if FindFilesWithPattern returns STATUS_NOT_IMPLEMENTED

accorp:
Small enhancement to FindFiles without breaking API compatibility.

### DIFF
--- a/dokan/directory.c
+++ b/dokan/directory.c
@@ -457,7 +457,12 @@ VOID DispatchDirectoryInformation(HANDLE Handle, PEVENT_CONTEXT EventContext,
           EventContext->Operation.Directory.DirectoryName, pattern,
           DokanFillFileData, &fileInfo);
 
-    } else if (DokanInstance->DokanOperations->FindFiles) {
+    } else {
+      status = STATUS_NOT_IMPLEMENTED;
+    }
+
+    if (status == STATUS_NOT_IMPLEMENTED &&
+        DokanInstance->DokanOperations->FindFiles) {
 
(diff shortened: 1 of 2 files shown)


Liryna:
Hi @accorp,

I understand in which case this fix could be useful but,
is it not 'strange' that the FindFilesWithPattern would be implemented and return STATUS_NOT_IMPLEMENTED ?

accorp:
Not as strange as to give callbacks possibility to return result as NTSTATUS and disallow one of its proper code. :smile: BTW In Dokan.NET binding there is NotImplemented as one of DokanResult constant.

This issue I faced trying to write full featured class-based dokan wrapper for Delphi/Lazarus. Currently, FindFilesWithPattern hides FindFiles if it is set in DOKAN_OPERATIONS. If all callbacks will be able to return STATUS_NOT_IMPLEMENTED it gives control from designtime to runtime, which is required for high-level programming. I think it will be useful for other bindings as well.

As a possible workaround, I can use separate DokanOperations for each mount, or make call to FindFiles from FindFilesWithPattern and then use DokanIsNameInExpression, or not support FindFilesWithPattern at all (like Dokan.NET binding does), but I don't like any of them.

Liryna:
Ok that make it more clean :+1: 
Thanks for the clear answer !

It even look more like how we use EnumerateNamedStreams in .NET as you say.

I have also add a ticket for the Dokan.NET wrapper FindFilesWithPattern that is missing for now.
https://github.com/dokan-dev/dokan-dotnet/issues/60

accorp:
Great, thanks!

PR #155, opened 2016-02-04
Support STATUS_NOT_IMPLEMENTED as a possible result from callbacks

accorp:
Addition to my previous PR. Handle STATUS_NOT_IMPLEMENTED exactly same as if the callback was not set in DokanOperations. Most of callbacks already can safely return it, this is intended to support for the rest.

### DIFF
--- a/dokan/flush.c
+++ b/dokan/flush.c
@@ -28,6 +28,7 @@ VOID DispatchFlush(HANDLE Handle, PEVENT_CONTEXT EventContext,
   PEVENT_INFORMATION eventInfo;
   ULONG sizeOfEventInfo = sizeof(EVENT_INFORMATION);
   PDOKAN_OPEN_INFO openInfo;
+  NTSTATUS status;
 
   CheckFileName(EventContext->Operation.Flush.FileName);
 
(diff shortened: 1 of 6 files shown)


accorp:
Also I found a suspicious place, but not absolutely sure to fix it myself.

[setfile.c, DokanSetAllocationInformation()](https://github.com/dokan-dev/dokany/blob/afaf3569e96d0da4f21a3141a93a3f3cb70f2daa/dokan/setfile.c#L52)

```c
  // How can we check the current end-of-file position?
  if (allocInfo->AllocationSize.QuadPart == 0) {
    return DokanOperations->SetEndOfFile(
        EventContext->Operation.SetFile.FileName,
        allocInfo->AllocationSize.QuadPart, FileInfo);
  } else {
    DbgPrint("  SetAllocationInformation %I64d, can't handle this parameter.\n",
             allocInfo->AllocationSize.QuadPart);
  }

  return STATUS_SUCCESS;
```

This code is a fallback to SetEndOfFile if SetAllocationSize is not implemented. I think it should be removed, it is application responsibility to implement required operations. Anyway, there is a potential crash if SetEndOfFile is not set.

Liryna:
Hi @accorp ,

Thank you again for your contributions !

You changes seems to be correct and I like the idea that all functions can return STATUS_NOT_IMPLEMENTED (even if should not :smile: ).

About DokanSetAllocationInformation, I agree with you, this should be rewrite. Can add it to your PR?

accorp:
@Liryna
I just removed that code, is it correct? Or it should return STATUS_SUCCES if SetAllocationSize not implemented as it was.

Liryna:
hi @accorp ,

Yes, I think this is the normal behaviour that should be for SetAllocation!

Thank you again for you contribution !

(I have failed my merge :cry: so I had to commit your changes myself https://github.com/dokan-dev/dokany/commit/a87dddd0e69051f99ce0d4c54efa7165f79ffc7b sorry for it)

PR #160, opened 2016-02-07
support hibernation mode

marinkobabic:
If system goes into hibernation mode it can happen that DokanCheckKeepAlive remove the MountPoint by mistake.

### DIFF
--- a/sys/timeout.c
+++ b/sys/timeout.c
@@ -260,6 +260,9 @@ Routine Description:
   LARGE_INTEGER timeout = {0};
   BOOLEAN waitObj = TRUE;
 
+  LARGE_INTEGER LastTime = { 0 };
+  LARGE_INTEGER CurrentTime = { 0 };
+
   DDbgPrint("==> DokanTimeoutThread\n");
 
   KeInitializeTimerEx(&timer, SynchronizationTimer);
@@ -268,6 +271,8 @@ Routine Description:
   pollevents[1] = (PVOID)&timer;
 
   KeSetTimerEx(&timer, timeout, DOKAN_CHECK_INTERVAL, NULL);
+
+  KeQuerySystemTime(&LastTime);
 
   while (waitObj) {
     status = KeWaitForMultipleObjects(2, pollevents, WaitAny, Executive,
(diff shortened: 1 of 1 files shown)


Liryna:
Hi @marinkobabic ,

Thank you for the pull request!

I never faced this issue but I can easily imagine what is happening. Your fix seems to be a good solution but I just would like to understand why the use of `((DOKAN_CHECK_INTERVAL + 2000) * 10000)` and not directly 1 minute (for ex) ?

marinkobabic:
Hi Liryna

You will face the problem when you use a notebook and you just close the screen. If your energy settings are set to suspend the machine when closing the screen, the notebook will go into the sleep mode. When you open the screen again, the machine will awake and run. Now if the DokanKeepAlive does the check the MointMoint will be deleted for sure. It may just happen sometimes. If as first the IOCL_KEEPALIVE comes in, all will work fine. It's a timing question. Repeat the step 10 times and you will face the problem.

DOKAN_CHECK_INTERVAL is set actually to 5 seconds. This means the timer will run every 5 second. If the machine goes into sleep mode the delay between the CurrentTime and LastTime will be more than 5 seconds. So I did 5 Seconds + 2 Seconds to be really sure, that it is a delay of timer. To use a value of 1 Minutes makes no sense because the timer can be configured using the constant DOKAN_CHECK_INTERVAL and I have no idea how long the machine is in hibernation mode. Every delay of more than 7 seconds means hibernation mode in my case. If the user is faster than 1 Minute, then the MountPoint will be deleted.

Liryna:
Alright thank for your clear answer !

But `((DOKAN_CHECK_INTERVAL + 2000) * 10000)` ->  ((5x1000 + 2000) x 10000) = 70000000 milliseconds -> 70000 seconds, am I wrong ? `* 10000` should be remove no ?

marinkobabic:
:smile:  Oh I see what you are talking about

https://msdn.microsoft.com/en-us/library/windows/hardware/ff553068(v=vs.85).aspx

KeQuerySystemTime - System time is a count of **100-nanosecond** intervals since January 1, 1601.

Liryna:
Ok, I see now :smile: 

Thank you @marinkobabic :))

PR #163, opened 2016-02-11
fix mounting issue

marinkobabic:
This should solve the remaining mounting issue and the redirector

### DIFF
--- a/sys/create.c
+++ b/sys/create.c
@@ -880,10 +880,7 @@ VOID DokanCompleteCreate(__in PIRP_ENTRY IrpEntry,
     DokanFreeFCB(fcb);
   }
 
-  irp->IoStatus.Status = status;
-  irp->IoStatus.Information = info;
-  IoCompleteRequest(irp, IO_NO_INCREMENT);
+  DokanCompleteIrpRequest(irp, status, info);
 
-  DokanPrintNTStatus(status);
   DDbgPrint("<== DokanCompleteCreate\n");
 }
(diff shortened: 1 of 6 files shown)


Liryna:
Hey @marinkobabic 

This work perfectly ! :relaxed: 

Your last contributions have really improved alot dokan and fixed the main issues ! :heart:

I will try to make a beta2 tomorrow with the new installer since the 1.0.0 have reached all milestones.

Maxhy:
I'm going to love this release, it will fix so many things excepted since years!

Thank you again @marinkobabic, I would have done the same for this network mount issue fix :smile:, good work.
Interesting things about boot sector you made here, I'm just not 100% sure about your File System Recognizer read request check here, wouldn't `if (fileObject == NULL && Irp->MdlAddress != NULL) {` match other cases?

marinkobabic:
It may be. We could also do a check how many bytes they are going to read. NTFS reads for example just the first 512 bytes. If we face some problems we can investigate.

Actually the Windows 8.1 default pdf viewer can't open pdf files. Oplock implementation seems to not be complete. 

If we once decide to support pnp this should be done in a separate bus driver.

The redirector implementation changed since Vista. Actual implementation feels like a hack just  to have a network drive. Redirector should be implemented in a separate dll. The drive should be a normal drive with redirector support. 

There are many things  to do :smiley:

Liryna:
@marinkobabic I agree with you there is a lot to do !
But we can already be proud for having reborn Dokan project and bring him at this level of device implementation !

As you seen, I have create different issues related to your todo list :smiley:
About pnp and redirector, could you create issues with somes informations ? (I don't really know what to do)
This would give opportunities to contributors to help us.

marinkobabic:
I totally agree and in generally I think you and Maxhy are doing a great job. Without you two, we wouldn't have dokan anymore. Thank you!

Let's go on :smile:

PR #167, opened 2016-02-12
Feature/fuse timeoutparameter

matepek:
Fetaure: Dokan has option to set timeout for operations. This wasn't available from Dokan FUSE library.
An option parameter "daemon_timeout=<sec>" added to Dokan FUSE library similar to osxfuse.


Liryna:
Hi @matepek ,

You are right, the fuse API was missing this option :smiley: !

PR #180, opened 2016-02-22
Fix:Saving fileinfo after file creation in fuse.

matepek:
The value of `fuse_file_info::fh` was `(uint64_t)(-1)`(<-- default value) after `fuse_create`.


matepek:
Hi @Liryna ,
It seems, some problem occured with the cygwin test, but not only in my pull request.
Let me know, if you think that, this is my bad. :)

Liryna:
Hi @matepek ,

It is a error that I have to fix dont worry :smile: 
I will look into your PR when I get the time  !

Liryna:
Well it seems to be correct :+1: 

Thank you again @matepek for you contributions !!! :trophy:

PR #181, opened 2016-02-22
Antivirus software tries to create files before we have some waiters

marinkobabic:
This is a workaround for this issue.


Liryna:
Hi @marinkobabic ,

Thank you for your contribution and for hunting this bug :trophy: !!!

PR #186, opened 2016-03-03
fix/saving MS Word files

matepek:
Trying to save a Microsoft Office Word file on DokanFuse drive isn't working.
In my case it writes a much bigger file. Like ~131 KB, but it should be around 11 KB.

It turned out that Word works like this on normal case:
- SetEndOfFileInformationFile (~131 KB)
- SetAllocationInformationFile (~131 KB)
- many WriteFile calls; writes the content which size is ~11 KB
- SetEndOfFileInformationFile (~11 KB); shrinks back the filesize to content's size
- SetAllocationInformationFile (~11 KB)

On the case of using DokanFuse the SetAllocationInformationFile's result is "NOT IMPLEMENTED", so it tries to
- SetEndOfFileInformationFile (~131KB) - OK
- SetAllocationInformationFile (~131KB) - NOT IMPLEMENTED

for a few times, then gives it up, and only write the files content, but not shrinks the filesize to the content's size, so there are left some garbage. (I think it shouldn't be a crucial thing to get a NOT_IPM. for SetAlloc., but it is for Office.)

In this branch, I implemented one more DokanOperation called SetAllocationSize.
According to https://msdn.microsoft.com/en-us/library/windows/hardware/ff540232(v=vs.85).aspx, this function should shrink the file if it is bigger then the allocation size.

This fix hopefully won't mess up or change the current working, just makes Office to works.

Best Regards,

### DIFF
--- a/dokan_fuse/src/docanfuse.cpp
+++ b/dokan_fuse/src/docanfuse.cpp
@@ -325,6 +325,41 @@ static NTSTATUS DOKAN_CALLBACK FuseSetEndOfFile(
       impl->set_end_of_file(FileName, ByteOffset, DokanFileInfo));
 }
 
+static NTSTATUS DOKAN_CALLBACK FuseSetAllocationSize(
+  LPCWSTR FileName, LONGLONG ByteOffset, PDOKAN_FILE_INFO DokanFileInfo) {
+  impl_fuse_context *impl = the_impl;
+  if (impl->debug())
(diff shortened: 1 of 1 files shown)


Liryna:
Hi @matepek ,

Thank you for your contribution, very appreciate!

I see well the behaviour you describe here. I guess there is no equivalent function in FUSE ?

matepek:
Hi @Liryna ,
It really seem to me, there are not.

Liryna:
Ok :+1:  was just to be sure.

Thank you again for your great work !

PR #188, opened 2016-03-03
SetAssemblyVersion added

js69:
Tool to update version.xml file. 

Instructions for usage: edit version.txt to increase build revision; for a new release version copy old lines to top, keep history and changelogs on bottom.

Tool wil be built by build.bat and called before WiX setup is created:
```
REM set version info, edit version.txt before running the batch
if NOT exist SetAssemblyVersion\bin\Release\SetAssemblyVersion.exe (
	msbuild SetAssemblyVersion.sln /p:Configuration=Release /p:Platform="Any CPU" /t:rebuild 
)
SetAssemblyVersion\bin\Release\SetAssemblyVersion version.txt /setversion version.xml
 
```

Contents of version.txt will have to look like this (version info and build date is searched by the tool in the first 5 lines):
```
---------------------------------
v1.0.0.100 (RC2)
Build date: 20160303
---------------------------------
Changes:
- added something
---------------------------------
v1.0.0.99 (RC1)
Build date: 20160303
---------------------------------
Changes:
- first release with revision number
```

The tool will update the ProductCode lines in version.xml in format ProductCode="{65A3A964-3DC3-mjmi-buld-yyMMddHHmmss}" where mj = major version, mi=minor version, buld=build number followed by build date as set in version.txt. Revison number will be used for MSI and Bundle.

Never change the UpgradeCode!


Liryna:
Thank you @js69 for you contribution !

Thats is indeed something we need :smile: !

I will test it as soon as I can :+1: 

Otherwise, do you have an idea how to check well that an already dokan version with the same major API version is installed ?

js69:
When the version numbers (major/minor/build/revision) between MSI packages are different, MSI will take care of the upgrade process. That means the upgarde will first uninstall the old driver and should request a reboot before the new driver is installed.

For our purposes where we might encounter any random version of dokan.sys or dokanXYZ.sys we might want to do our own checks in a CustomAction.

Our DOKANSYSALREADYINSTALLED property check will currently fire only when we still find a dokan1.sys.

Liryna:
Ok good to know !

DOKANSYSALREADYINSTALLED is 'working' but when dokan driver install fail, it can happen that dokan1.sys is not removed...so you will get an error that can only be fixed if you removed it by hand.

If what you describe is working, we do not need DOKANSYSALREADYINSTALLED. I am going to remove it.

About dokan fork, they will have to change driver name and such to work in parallelle of dokan.

Liryna:
It seems you have remove that .cat from the installer in your last commit https://github.com/js69/dokany/commit/686744343aea93f0dc025134af8fdc0b3b20bb78#diff-29b04a548dd83489fcf856ec411749aeL168

js69:
Sorry, fixed now.

js69:
Latest change:

 Disallow upgrades, show uninstall and reboot message instead

Because MSI removes Dokan1.sys driver file only after reboot, an
automatic upgrade is not possible. We show an appropriate message
instead.

A custom bundle should detect an existing Dokan install and warn accordingly.

Liryna:
Hi @js69 ,

I tested it and it seems to work very well :+1: Well done it gonna be a great help !

I just have a question about https://github.com/dokan-dev/dokany/pull/188/files#diff-29b04a548dd83489fcf856ec411749aeR28
I have this during the build `This product should remove only o
lder versions of itself. The Maximum version is not less than the current product. (1.x.x.100 1.x.x.100)`
Does the restriction work like that:
 - If any dokan version with API 1 is installed -> it propose to uninstall/install ? 
 - If you have a Dokan 1 installer and Dokan API 2 already installed -> it accept to install Dokan 1 ?

If not, do you think you could include it in your PR ? :heart: 

(Also rebase all commit in one)

js69:
Currently all direct upgrades are disallowed, the MSI will show the following message when it detects a previous Dokan installation (with the same UpgradeCode):

"A previous version of [ProductName] is already installed. Please uninstall the existing version first, reboot the system and start the setup again."

There seems to be no safe way currently to do a direct DiFX upgrade when an intermittent reboot is required as it is for dokan*.sys. I've asked the WiX authors for some advice on that but it seems to be a limitation of the DiFX tech.

On your last question: you can ignore the build warning message. It would only be relevant when you run a setup when the identical version is already installed, allowing effectively a reinstall/repair upgrade in that case. But as we disallow all upgrades, it doesn't matter.

A completely different Dokan API driver, e.g. a Dokan2.sys, should use a different UpgradeCode. It will then be ignored and you can then run dokan1 and dokan2 side by side (as long as no conflicts exist, e.g. service name etc is different).

We can suppress that warning by adding -sw1076 in the tool settings, I've committed that change now.

Liryna:
Ok clear to me !  :relaxed:

A little rebase of your commit and I merge it :heart:

Liryna:
Here is how to rebase all your commit into one:
```
git reset --soft HEAD~7
git commit
```

js69:
Hmm, I would need to update to latest master in my repo, how would I best resolve conflicts in the git shell?

Liryna:
I will merge your pullrequest myself.

I will tag you in the commit message if thats alright for you.

js69:
Sure, no worries.

Liryna:
Merged with https://github.com/dokan-dev/dokany/commit/15dfb9a3549891d5ff54cfc9a45c7617997f7c70

Thank you @js69 for your contributions ! very appreciate ! :trophy:

PR #193, opened 2016-03-09
Use control handler function to gracefully close mirror sample on Ctrl+C

accorp:


Liryna:
:+1: This had to be done long time before.

Thank you @accorp !!!

accorp:
I run mirror under debugger, and it shows that exception is "read of address 0x00000000" somewhere inside ole32.dll. According to [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683242%28v=vs.85%29.aspx) when Ctrl+C is pressed, default control handler calls ExitProcess(), and it looks like some function from ole32 is not happy to be called after it. I got same crash with direct call to ExitProcess() while volume is mounted.

Anyway, it seems that call to SHChangeNotify() is useless, maybe it should be removed as you was planning to do.

Liryna:
Thank you for your testing @accorp .
I agree with your observations !

Therefore I will remove it. I just don't know that seems on Win 8/10 it seems that there is no crash, maybe I should just disable it for Win7 ?
What do you think about it ? Have you also only faced the issue on Win7 ?

accorp:
Crash on both Win7 x86 and x64, but not on Win8.1 x86. And I still can't reproduce it on win7 with delphi mirror even with direct call to ExitProcess().

Liryna:
Does you delphi run in a shell as the mirror does ?

accorp:
Yes, line-to-line translation of C version. But "delphi" (i'm actualy using lazarus) has some runtime compiled into executable and it does additional tasks for initialization/finalization/setting callbacks etc. And maybe one of them is fortunately avoid that crash.

Liryna:
Yes It has to be a story of timing/call order during the uninit.
I have remove it for Win7 https://github.com/dokan-dev/dokany/commit/df8e4d5a2dc31c2913c9be133c31bd6784f2aec8

Thank you again @accorp for your input, very appreciate !

accorp:
Hi @Liryna, I found the reason why delphi mirror is working properly - there is ole32.dll in import table. And I got no more crash after putting dummy call to CoGetCurrentProcess() inside dokan library just to add ole32 to import. But it looks like call to SHChangeNotify() during DLL_PROCESS_DETACH is a bad idea, see [DllMain entry point](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682583%28v=vs.85%29.aspx) and [Dynamic-Link Library Best Practices](https://msdn.microsoft.com/en-us/library/dn633971.ASPX).

"Calling functions that require DLLs other than Kernel32.dll may result in problems that are difficult to diagnose. For example, calling User, Shell, and COM functions can cause access violation errors, because some functions load other system components. Conversely, calling functions such as these during termination can cause access violation errors because the corresponding component may already have been unloaded or uninitialized."

Liryna:
Hi @accorp ,

So If I understood well, even by adding ole32.dll in import table, it is possible that it get unload before we get the change to call SHChangeNotify ?
Or does adding ole32.dll in import table 'force' to unload at the end of DLL_PROCESS_DETACH ?

Liryna:
I begin to think that we should just remove it during remove. We are already lucky to not have the crash in other systems.
And It is so much badly documented....that I cannot say if BroadcastSystemMessage does not already to the SHChangeNotify job.

accorp:
I agree with your decision, looks like adding ole32 into import table keeps it from unloading too fast and solved the problem in this particular case. But it is not guarantee to remain working, at different configurations or after some updates it can start crashing again.

Liryna:
Done with https://github.com/dokan-dev/dokany/commit/cebf6ca1fc4e2b3ed2f7204d5e62b4d4cc992187.

If we see users having issue like the device is not removed for explorer or such, we will try to find another way to notify.

PR #194, opened 2016-03-10
remove some compiler complaints

x1596357:
gcc in msys2 always complains about implicit pointer conversions and treats them as errors.


Liryna:
Hi @x1596357 ,

Thank you for your contribution !

Good luck for vramfs-on-Windows and feel free to ask for adding it to dokan homepage when it is read ^^

PR #195, opened 2016-03-11
use struct constructor in fuse example

x1596357:
When fusexmp.c is compiled using msys2 toolchain, it would prompt "sorry, unimplemented: non-trivial designated initializers not supported". By using struct constructor, fusexmp.c can be compiled in both msys2 and cygwin.

### DIFF
--- a/samples/fuse_mirror/fusexmp.c
+++ b/samples/fuse_mirror/fusexmp.c
@@ -382,43 +382,45 @@ static int xmp_removexattr(const char *path, const char *name)
 }
 #endif /* HAVE_SETXATTR */
 
-static struct fuse_operations xmp_oper = {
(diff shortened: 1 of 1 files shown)


Liryna:
Oh ! you succeed to make it build totally ?
Does it run properly ?

I had some issues for building it with msys2 also and had no time to look into it.

x1596357:
Successfully building it both with msys2 and cygwin. I tested running in msys2 and it works.  However there remains some issues not related to this PR, like all file size showed in explorer is 0 but in cmd it is not. I think they are related to the  structs used in dokanfuse.

x1596357:
breaks building in Appveyor, I am gonna check why.

PR #209, opened 2016-03-31
use _WIN32 in FUSE wrapper headers instead of WIN32

x1596357:
We should use _WIN32 instead WIN32 in FUSE wrapper headers. _WIN32 is safer. 
And more importantly when using these wrapper headers to develop applications, WIN32 is not defined in` fuse.h#L13 `when in x64 platform. `fuse_win.h` needs to be included for later(`fuse.h#L448`) WIN32 being defined.


Liryna:
Thank you @x1596357 for your contribution :+1: !

PR #211, opened 2016-04-11
Fuse handles GetFileSecurity callbacks.

matepek:
Creating a new folder with explorer under a sub-folder of the mounted path shows it needs higher privileges.

needpriv

If we fill it, it will work.
This pull request queries the privileges of the mount path, gives back to every request which occurs.
It was written based on mirror.c.

### DIFF
--- a/dokan_fuse/src/dokanfuse.cpp
+++ b/dokan_fuse/src/dokanfuse.cpp
@@ -433,6 +433,101 @@ static NTSTATUS DOKAN_CALLBACK FuseUnmounted(PDOKAN_FILE_INFO DokanFileInfo) {
   return errno_to_ntstatus_error(impl->unmounted(DokanFileInfo));
 }
 
+static DWORD AddSeSecurityNamePrivilege() {
+    LUID luid;
+    if (!LookupPrivilegeValue(0, SE_SECURITY_NAME, &luid)) {
+        DWORD err = GetLastError();
+        if (err != ERROR_SUCCESS) {
+            return err;
+        }
+    }
+
+    HANDLE token = 0;
+    if (!OpenProcessToken(GetCurrentProcess(),
+        TOKEN_ADJUST_PRIVILEGES | TOKEN_QUERY, &token)) {
+        return GetLastError();
+    }
+
+    ON_BLOCK_EXIT([token] {
+        CloseHandle(token);
+    });
+
+    LUID_AND_ATTRIBUTES attr;
+    attr.Attributes = SE_PRIVILEGE_ENABLED;
+    attr.Luid = luid;
+
+    TOKEN_PRIVILEGES priv;
+    priv.PrivilegeCount = 1;
+    priv.Privileges[0] = attr;
+
+    AdjustTokenPrivileges(token, FALSE, &priv, NULL, NULL, NULL);
+    DWORD err = GetLastError();
+    if (err != ERROR_SUCCESS) {
+        return err;
+    }
+
+    return ERROR_SUCCESS;
+}
+
+static NTSTATUS DOKAN_CALLBACK FuseGetFileSecurity(
+    LPCWSTR FileName, PSECURITY_INFORMATION SecurityInformation,
+    PSECURITY_DESCRIPTOR SecurityDescriptor, ULONG BufferLength,
+    PULONG LengthNeeded, PDOKAN_FILE_INFO DokanFileInfo) {
+  impl_fuse_context *impl = the_impl;
+  if (impl->debug())
+    FWPRINTF(stderr, L"GetFileSecurity %x\n", *SecurityInformation);
+
+  BY_HANDLE_FILE_INFORMATION byHandleFileInfo;
+  ZeroMemory(&byHandleFileInfo, sizeof(BY_HANDLE_FILE_INFORMATION));
+
+  int ret;
+  {
+      impl_chain_guard guard(impl, DokanFileInfo->ProcessId);
+      ret = impl->get_file_information(FileName, &byHandleFileInfo, DokanFileInfo);
+  }
+
+  if (0 != ret) {
+    return errno_to_ntstatus_error(ret);
+  }
+
+  if (byHandleFileInfo.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY) {
+    // We handle directories for the Explorer's 
+    // context menu. (New Folder, ...)
+
+    HANDLE handle = CreateFileW(
+        DokanFileInfo->DokanOptions->MountPoint,
+        READ_CONTROL | (((*SecurityInformation & SACL_SECURITY_INFORMATION) ||
+        (*SecurityInformation & BACKUP_SECURITY_INFORMATION)) 
+            ? ACCESS_SYSTEM_SECURITY : 0),
+        FILE_SHARE_WRITE | FILE_SHARE_READ | FILE_SHARE_DELETE,
+        NULL, // security attribute
+        OPEN_EXISTING,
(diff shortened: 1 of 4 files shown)


Liryna:
Hi @matepek ,

I totally understand the behaviour you describe and your implementation seems to do the job :+1: 

It seems that there is some build issue with cygwin & msys.

```
In file included from /cygdrive/c/projects/dokany/dokan_fuse/src/dokanfuse.cpp:3:0:
/cygdrive/c/projects/dokany/dokan_fuse/include/fusemain.h: In constructor ΓÇÿwin_error::win_error(int)ΓÇÖ:
/cygdrive/c/projects/dokany/dokan_fuse/include/fusemain.h:58:55: error: ΓÇÿerrno_to_ntstatus_errorΓÇÖ was not declared in this scope
  win_error(int _err): err(errno_to_ntstatus_error(_err)) {}
                                                       ^
 
/cygdrive/c/projects/dokany/dokan_fuse/src/dokanfuse.cpp: In function ΓÇÿNTSTATUS FuseFindFiles(LPCWSTR, PFillFindData, PDOKAN_FILE_INFO)ΓÇÖ:
/cygdrive/c/projects/dokany/dokan_fuse/src/dokanfuse.cpp:47:62: error: ΓÇÿerrno_to_ntstatus_errorΓÇÖ was not declared in this scope
       impl->find_files(FileName, FillFindData, DokanFileInfo));
                                                              ^
 
/cygdrive/c/projects/dokany/dokan_fuse/src/dokanfuse.cpp: In function ΓÇÿNTSTATUS FuseDeleteDirectory(LPCWSTR, PDOKAN_FILE_INFO)ΓÇÖ:
/cygdrive/c/projects/dokany/dokan_fuse/src/dokanfuse.cpp:68:54: error: ΓÇÿerrno_to_ntstatus_errorΓÇÖ was not declared in this scope
       impl->delete_directory(FileName, DokanFileInfo));
                                                      ^
 
 ....
```

Also have you seen in the mirror that in some case you will need admin right and request SE_SECURITY_NAME ?
https://github.com/dokan-dev/dokany/blob/master/samples/dokan_mirror/mirror.c#L1393

Liryna:
Oh sorry @matepek , the build error are not related to your PR :smile:

dokan_fuse/src/dokanfuse.cpp
```diff
@@ -433,6 +433,87 @@ static NTSTATUS DOKAN_CALLBACK FuseUnmounted(PDOKAN_FILE_INFO DokanFileInfo) {
...
+    PULONG LengthNeeded, PDOKAN_FILE_INFO DokanFileInfo) {
+  impl_fuse_context *impl = the_impl;
+  if (impl->debug())
+    FWPRINTF(stderr, L"GetFileSecurity %x\n", *SecurityInformation);
+
+  bool needASS = ((*SecurityInformation & SACL_SECURITY_INFORMATION) ||
+      (*SecurityInformation & BACKUP_SECURITY_INFORMATION));
+
+  if (needASS) {
+      DWORD result = AddSeSecurityNamePrivilege();
```
matepek: Is it OK to call it anytime it occurs? I mean (it seems) WinAPI doesn't mind, but maybe it is a performance issue.
What do you think? Should I rewrite it to be called only once?

matepek:
Hello @Liryna,

You are right, I missed that part, ty. :)
I updated the pull request.
Please review it carefully.
There is a question on the diff already. :)

I choose this solution, to call *AddSeSecurityNamePrivilege* only if it is necessary, because we don't want more privilege from the user. So according this solution, the *FuseGetFileSecurity* can fail with *STATUS_NOT_ALL_ASSIGNED*. It seems acceptable. are you agree?

dokan_fuse/src/dokanfuse.cpp
Liryna: @matepek does this should not be made in fuse_main_real ?
Like that it would only be made one time.

Also if AddSeSecurityNamePrivilege fail, it is better to ignore it. FuseGetFileSecurity could still work in different case.

dokan_fuse/src/dokanfuse.cpp
```diff
@@ -433,6 +433,87 @@ static NTSTATUS DOKAN_CALLBACK FuseUnmounted(PDOKAN_FILE_INFO DokanFileInfo) {
...
+      if (ERROR_SUCCESS != result)
+          return lasterror_to_ntstatus(result);
+  }
+
+  HANDLE handle = CreateFileW(
+      DokanFileInfo->DokanOptions->MountPoint,
```
Liryna: @matepek Oh this seems to be a dirty hack :smile: I am not sure the system is not going to like to get everytime the same security informations.

Your implementation would be the same as returning virtual static security information since `DokanFileInfo->DokanOptions->MountPoint` as been created by dokan.

I understand why you did this since you cannot request the informations to the file system or by the fuse API.
But I am sceptical on this implementation :cry:

matepek: Hello @Liryna ,
I must say, I tried to alloc a static memory which was filled with
GetUserObjectSecurity(MountPath, ...), and every time when FuseGetFileSecurity call occurs, memcpy the content from it to the SecurityDescriptor, but I couldn't make it to work yet.
There are known lack in my Windows Security knowledge, I welcome any help. :)
I modified mirror.c which is now produce the same "bug".
http://pastebin.com/HgS2tu8u (it's ugly, it is for proof of concept)
Could you look on it?

Liryna: Hello @matepek ,

I was meaning that FuseGetFileSecurity is always going to return the same informations since it always open the same file when he should return the specific information related to the file requested.

Therefore, I am not sure the system will like to get informations of a directory when he request a file.
Maybe the implementation should return different values when it is a file, directory,...requested ?

I am also not really familiar with SECURITY_INFORMATION :cry: sorry.
https://msdn.microsoft.com/fr-fr/library/windows/desktop/aa379573%28v=vs.85%29.aspx

matepek:
Hello @Liryna,

I modified the source.
Do you agree now?
The files are not handled, only directories.

But, an idea just came. We can create a temp file into the user's temp directory, and read the security privileges from it. What do you think about it?

P.S.: Do you want me to compress, the branch history into one commit, as previously?

Liryna:
Hi @matepek ,

Yes this should do the work thank you for the changes ! :+1:
Does this still make your case work ?

Unfortunately, I don't have access to my env for the moment :'( I will merge it when I can.

A clean history would be welcome yes !

matepek:
Hi @Liryna ,

Yes, it seems, this will do for us.
History has been cleared.

Liryna:
Hi @matepek ,

Sorry for the delay !
Everything seems to be alright for me !

I just have a question not related to the PR, care you still able to build the fuse mirror sample and run it correctly ? are you able to read / write files also ?
I have some issue on my side so it would be nice if someone could test it :( I hope there is no regression.

matepek:
Hello @Liryna ,
Do you mean ~~fuse~~ mirror sample?
Yess, I also can feel the disturbance in the force.
I haven't enough time to debug it, but I can say:

"mirror.exe /r c:\Users\mate.pek\mirror_it\ /l e /d" as administrator.
could rw: .txt with notepad++;
could open: .xls (w not tested)
couldn't open: .docx; .pdf

I hope I helped a bit.

Liryna:
I was talking about the **fuse** mirror in case you had cygwin :smiley: 
but thats alright I will try to do the test on my side again

matepek:
Oh, totally missed that. ^^

PR #214, opened 2016-04-20
networkdrive fileObject has no Vpb

marinkobabic:
Fix for REFERENCE_BY_POINTER when NETWORK_DRIVE is used


Liryna:
Hi @marinkobabic ,

Sorry for the delay, I have finally been able to review the PR with @Maxhy .

Your fix work perfectly even if we dont really why this is happening and there is not documentation that seems to give informations about it :cry: 

Thanks again for your contribution it is always very appreciate !

PR #217, opened 2016-04-26
dokan: null-terminate string in unmount

taruti:
Spotted this bug while hunting other issues.

This may have caused invalid input to DeleteVolumeMountPoint in the !IsMountPointDriveLetter case.


Liryna:
Hi @taruti ,

Well found !

Thank you for you contribution !

PR #218, opened 2016-04-27
FUSE: Fix capitalization to enable cross-compiling from Linux

Rondom:
I fixed two small issues where the capitalization of filenames differed from what was expected. This broke compilation on case-sensitive filesystems i.e. when cross-compiling on Linux.


Liryna:
Hi @Rondom ,

You are right :) I am only building on Windows so never had such issue.

Thank you for your contribution !

Issue #236, opened 2016-05-12
LibDokan RC3 fails to open some files


psaghelyi:
# Environment
 * Windows version: Windows 8.1
 * Processor architecture: x86 (core i7)
 * Dokany version: 1.0.0 RC2
 * Library type (Dokany/FUSE): dokany
 
# Check List
  - [X] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [X] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [X] I can always reproduce the issue with the provided description below.
  - [X] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description
In Dokan RC2 I could open-edit-save MSWord documents. In Dokan RC3 I cannot.
sample: [01-HttpCient-Es-A-Logolas.docx](https://github.com/dokan-dev/dokany/files/261250/01-HttpCient-Es-A-Logolas.docx)

Symptom: Open MSWord doc using WindowsExplorer and TotalCommander. They freeze then crash after certain seconds.

# Logs
[kernel.zip](https://github.com/dokan-dev/dokany/files/263341/kernel.zip)
[mirror.zip](https://github.com/dokan-dev/dokany/files/263342/mirror.zip)

Liryna:
Hi @psaghelyi ,

It is not easy and quick to reproduce like you say since every MSWord behave differently.
We will need mirror and kernel logs here.

psaghelyi:
Some refinement:
- It has the same problem on Windows10 and Windows8.1.
- You can open/edit/save the Word document if you open them from MSWord2016
- Only Explorer and TotalCommander crashes on open attempt.
- Tested on different machines, clean install
- please find the logs in the attachment
[kernel.zip](https://github.com/dokan-dev/dokany/files/263315/kernel.zip)
[mirror.zip](https://github.com/dokan-dev/dokany/files/263316/mirror.zip)

Liryna:
Hi @psaghelyi ,

I have made the test windows 10 x64 and Word 2016.
I had no issue by using your file with Explorer / Total Commander or directly from Word.

`The CreateFile was not even called so probably the root of the problem is deeper in Dokany.sys.`
I see a couple of CreateFile in your logs but there is no ReadFile request.

Can you remake the test without having other files in the same folders ?

Maxhy:
[...]

Issue #234, opened 2016-05-11
FUSE stat attributes wrong


gang929:
In dokan fuse, when readdir is called, dokanfuse use the stat for fill file&dir attr, but stat define is depends platform. 
For example,  the mingw stat define is different to linux,  so, when compile app with dokanfuse, must use the dokanfuse compile platform!
In my wdfs port project, I use mingw compile, so the dir, file length, file create date, file modify date is not correctly convert and display.
Have any way to solve this problem?

Liryna:
`the mingw stat define is different to linux` could you link this part of code in the dokan fuse ?
`In my wdfs port project, I use mingw compile` So wdfs is using mingw stat at this moment ? and dokan fuse build with mingw does not ?

gang929:
in wdfs' readdir interface， as below, the filler will use  file stat  to fill
```
wdfs_readdir(const char *localpath, void *buf, fuse_fill_dir_t filler, off_t offset, struct fuse_file_info *fi)
```
in mingw, the stat define as below 
```
 struct stat {
    _dev_t st_dev;
    _ino_t st_ino;
    unsigned short st_mode;
    short st_nlink;
    short st_uid;
    short st_gid;
    _dev_t st_rdev;
    _off_t st_size;
    time_t st_atime;
    time_t st_mtime;
    time_t st_ctime;
  };
```
but in dokanfuse, then function  convertStatlikeBuf in utils.h use stat maybe as below (depends on dokanfuse compile env)
```
struct stat {
.......
    struct timespec st_atim;            /* Time of last access.  */
    struct timespec st_mtim;            /* Time of last modification.  */
    struct timespec st_ctim;            /* Time of last status change.  */
.......
}
```
It seems that is not binary compatible.

Liryna:
Ok I see, so it would mean that:
https://github.com/dokan-dev/dokany/blame/f5afbea312cf69fcd217cb14fb2af51067d89b83/dokan_fuse/include/fuse_win.h#L74

should be changed to: `typedef struct timespec time_t;` ?

gang929:
NO, this is a  binary compatible problem, only changing the st_xxxx define  can not solve it.
In order to solve this problem，It should Export all struct used by Dokanfuse and App developed with Dokanfuse.
For example, the stat struct can be export as FUSE_STAT, When other App developing  with dokanfuse, it should convert it's stat to FUSE_STAT,  to ensure the binary compatibility of stat struct.

gang929:
[...]

PR #242, opened 2016-05-17
fuse:fix stat define and add network drive map

gang929:
use volname for network drive name
add option use network drive
redefine stat for binary compatible

### DIFF
--- a/dokan_fuse/include/dokanfuse.h
+++ b/dokan_fuse/include/dokanfuse.h
@@ -15,6 +15,7 @@ struct fuse_config
     int debug;
 	int setsignals;
     unsigned int timeoutInSec;
+	int useNetworkDrive;
 };
 
 struct fuse_session
--- a/dokan_fuse/include/fuse_win.h
+++ b/dokan_fuse/include/fuse_win.h
@@ -8,7 +8,7 @@
 #include <time.h>
 #include <sys/types.h>
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || defined(__MINGW32__)
 #include <windows.h>
 #endif
 
@@ -74,6 +74,7 @@ typedef unsigned __int64 fsblkcnt64_t;
 typedef struct timespec timestruc_t;
 typedef unsigned short nlink_t;
 typedef unsigned __int64 uint64_t;
+typedef unsigned __int32 uint32_t;
 typedef unsigned int blksize_t;
 typedef unsigned __int64 blkcnt_t;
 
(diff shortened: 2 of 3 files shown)


Maxhy:
Two remarks:
 - I think you're setting volume name from registry because you don't want to use UNC here or didn't registered the Network Provider (there is indeed a small today, to setup a custom volume name on network mount point, you have to setup a UNC redirector)? If so, this shouldn't be specific to FUSE but generally available on Dokan Library. But I also believe it would be better to refactor UNC redirector setup logic to avoid to use registry and duplication with Network Provider (be aware of current bug https://github.com/dokan-dev/dokany/issues/236#issuecomment-219688455). What do you think?
 - When your PR is ready, is better to clean git history to one commit per change compared to base (see https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/).

Thanks @gang929.

gang929:
Because i want use the command line parameter to set the Network Map Drive Name each time, so use the parameter volame for the network drive name. And, I didn't find the code to set the network drive name("DAV RPC SERVICE")

Maxhy:
I updated the wiki to explain better how a network drive label could be setup: https://github.com/dokan-dev/dokany/wiki/API#network-provider
I agree your code to setup from registry could have its interest, but it shouldn't be on fuse library only then.

Otherwise, I'm not sure to understand your binary compatibility change here. I know it is related to #234 but could you explain what is in details the issue and how your changes are fixing it here? I don't see the value of these kind of defines here, probably just a lack of knowledge from my side. Thanks.

gang929:
sorry, I  only know very little about network-provider,  I didn't know how to set an UNC Name for Network Redirector, so ,I directly setup drive label from registry. 
the `dokanctl.exe /i n` code can not set unc name with mapping a drive each time 
 @Maxhy , you can explain more about?  thanks!

about https://github.com/dokan-dev/dokany/issues/234 
In VS2015(build with x86), the mode_t and nlink_t is 4bytes, but In mingw32, they are 2bytes,  otherwise, the timespec is 16bytes in VS2015, but 8bytes in mingw.  Maybe, the cygwin is same as VS2015.
so, I think the stat should be defined by fixed bytes for all field.

gang929:
the command `C:\Program Files\Dokan\Dokan Library-1.0.0\x86>sample\mirror\mirror.exe /r c: /l k /n` 
will mirror c: to k:, but the label sometimes is DOKAN,
when mirror to other driveletter(I, L, M ...)  sometimes is other unkown name (cg DAV RPC SERVICE， or xx\adfs\cvv)， it seems that this is a  bug?

Liryna:
@gang929 UNC can be set at the mount: https://github.com/dokan-dev/dokany/blob/master/samples/dokan_mirror/mirror.c#L1335

About label that sometimes change, can you make a test with the last commit: https://github.com/dokan-dev/dokany/commit/d3ecfb0e5d327092113430bfd7b27ee605860081
It can probably fix the behaviour. If not, can you describe a way to reproduce it ?

For binary compatibility, I don't understand why you compare visual studio and cygwin. The way it work, if your filesystem is build with cygwin, it has to use the dokan fuse build with cygwin. That's why the dokan installer has VS/Cygwin/msys version of the dokan fuse. With this you will have the same size of stat member no ?

gang929:
I rebuild  with the last commit:  https://github.com/dokan-dev/dokany/commit/d3ecfb0e5d327092113430bfd7b27ee605860081  and test 
`C:\Program Files\Dokan\Dokan Library-1.0.0\x86>sample\mirror\mirror.exe /r c: /l m /n /u ffff  `
in windows explorer, it show as 


For binary compatibility,  the dokanfuse1.dll build with VS2015,  I build wdfs with dokanfuse1.dll in mingw32.
if define fuse_stat.st_mode/st_nlink/st_atime... as below 
```
     mode_t        st_mode;
     nlink_t       st_nlink;
...
    timestruc_t   st_atim;
```
so , the sizeof(fuse_stat)  in dokanfuse1.dll  not equal sizeof(fuse_stat) in wdfs, because In VS2015(build with x86), the mode_t and nlink_t is 4bytes, but In mingw32, they are 2bytes, otherwise, the timespec is 16bytes in VS2015, but 8bytes in mingw.

gang929:
@Liryna  how to set unc name or network drive label?

Liryna:
@gang929 

UNC:

> mirror.exe /r c:\Users /l m /n /u \localhost\mytest

You can acces it by `\\localhost\mytest`

Network drive label is set by `GetVolumeInformation`
https://github.com/dokan-dev/dokany/blob/master/samples/dokan_mirror/mirror.c#L1125

gang929:
@Liryna 
`GetVolumeInformation ` is only  correctly with `mirror.exe /r c: /l m`, when `mirror.exe /r c: /l m /n` is display error as

Liryna:
From my point of view, I don't think we should add network option in libfuse since it does not exist in libfuse.

I will try to reproduce your issue @gang929 with wrong label name using network.

Maxhy:
Ok according to the discussion I will summarize this PR state as following:

 - Network label name: current method to set network label name is using UNC and proposed registry method should be avoided for now. A buffer related bug seems to exist and should be fixed.
 - stat64_cygwin structure is incompatible between compilers but this shouldn't be changed. This struct is generally used by fuse fs implementation with system function (eg. lstat), those are also dependent to each compiler (cygwin, msys2, ...). Your fuse fs should link with dokanfuse built with the same compiler.
 - My opinion about network option on libfuse is that it should indeed be implemented as fuse_config structure doesn't impact fs implementation (and if using *Windows only fuse options* a #ifdef check could be done).

Liryna:
Hi @gang929 ,

I have import and clean your commit for network drive option: https://github.com/dokan-dev/dokany/commit/8e4590c2a4b93a8d45366660c54e64c66975c39d
Thank you again for your contribution and your time !

I have also try to reproduce the issue with the mirror network name :cry: but impossible to get the same behaviour as you. If you have new informations about it, please open an issue ! :heart:

PR #251, opened 2016-05-30
Call statfs with root directory /

Rondom:
Instead of calling statfs with an empty string we now call statfs with
the root directory "/". This is in line with POSIX which requires the
path to be valid and existing (an empty path is not)

I removed the comment `// Why do we need path argument??`. The answer to the question is that, on a Unix system, the path is needed to determine which filesystem you want to get the information from, i.e. depending on where the path resides, you would get the information about the filesystem the file/dir is located on.

http://linux.die.net/man/2/statfs


Liryna:
Hi @Rondom ,

Thank you for the contribution and the explanation :+1: very appreciate !

About the comment, I just think it is because the dev was meaning that anyway the path is related to the dokan filesystem.

PR #252, opened 2016-05-31
dokan_fuse: Use %ls for wide-character debug-print

Rondom:
In dokan_fuse debug output the filenames are now printed out using %ls
in order to be compliant with the C++ standard and thus make the
printing work when compiling with GCC/MingW (without the printing would stop after the first character, if the second byte is NULL)

MSVC will happily print a wide-character-string when using %s but will
fail to print out single-byte strings. *This has to be kept in mind
when printing out single-byte strings using wide-character fwprintf*.

MSVC14 supports defining _CRT_STDIO_ISO_WIDE_SPECIFIERS to bring their
format-string handling in line with the standards.

From what I have read, Microsoft's decision to be not in line with the standard was to simplify compiling the same code-base for both unicode- and non-unicode systems back in the days. They tried to change it with MSVC14, but that was too much of a headache it seems. So I guess that is why now one has to define a macro to enable standard-compliant behaviour.... oh well...

https://blogs.msdn.microsoft.com/vcblog/2014/06/18/c-runtime-crt-features-fixes-and-breaking-changes-in-visual-studio-14-ctp1/

*Please test a compile with MSVC if you can (the actual debug-prints, not only the compiling itself). I only tried a test-program to check that the behaviour is as described above on MSVC. I am lacking a driver using dokan_fuse that compiles on MSVC.*

### DIFF
--- a/dokan_fuse/src/dokanfuse.cpp
+++ b/dokan_fuse/src/dokanfuse.cpp
@@ -40,7 +40,7 @@ FuseFindFiles(LPCWSTR FileName,
               PDOKAN_FILE_INFO DokanFileInfo) {
   impl_fuse_context *impl = the_impl;
   if (impl->debug())
-    FWPRINTF(stderr, L"FindFiles :%s\n", FileName);
+    FWPRINTF(stderr, L"FindFiles :%ls\n", FileName);
 
   impl_chain_guard guard(impl, DokanFileInfo->ProcessId);
   return errno_to_ntstatus_error(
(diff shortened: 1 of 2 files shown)


Rondom:
This needs more investigation and is not ready yet.

Rondom:
I have now done some brief testing by using my MingW-compile FUSE driver against the MSVC-compiled dokanfuse.dll. The paths were printed correctly.

Maxhy:
Indeed this is much better.
Thank you for this fix, and thanks a lot for your good explanation :+1:.

PR #258, opened 2016-06-12
Fixes for VC2015Redist 23918

js69:
Fix incorrect downloadurl, fix comparison


Liryna:
Well found @js69  :heart: !!!
We plan to make the release 1.0.0 next week so this is very helpful :D !

Issue #244, opened 2016-05-19
Can not install in windows server 2008 R2


superhq:
# Environment
 * Windows version: Windows Server 2008 R2 Enterprise
 * Processor architecture: 64 bit Intel(R) Core(TM) i5-3317U 
 * Dokany version: 1.0.0-RC3
 * Library type (Dokany/FUSE): Dokany
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [ ] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description
I run DokanSetup.exe in WS-2008-R2, everything seems find. But the mirror sample can not work,and output error message:

> Can't install driver

# Logs
I try to reinstall the driver use dokanctl.exe, and the log as follow:

> PS C:\Program Files\Dokan\Dokan Library-1.0.0> .\dokanctl.exe /i  d
Driver path: 'C:\Windows\system32\drivers\dokan1.sys'
Installing driver...
DokanServiceInstall: Service (Dokan1) is already installed
Driver install failed
PS C:\Program Files\Dokan\Dokan Library-1.0.0> .\dokanctl.exe /r d
Driver path: 'C:\Windows\system32\drivers\dokan1.sys'
Removing 'Dokan1'...
DokanServiceControl: Service (Dokan1) deleted
'Dokan1' removed.
PS C:\Program Files\Dokan\Dokan Library-1.0.0> .\dokanctl.exe /i d
Driver path: 'C:\Windows\system32\drivers\dokan1.sys'
Installing driver...
DokanServiceInstall: Service (Dokan1) installed
DokanServiceControl: Failed to start service (Dokan1). error = 577
DokanServiceInstall: Service (Dokan1) start failed
Driver install failed

I try to uninstall the driver,restart OS and reinstall the driver, but the result is same.

Liryna:
Hi @superhq ,

Error 557 mean that the system does not recognize the driver signature.
You will need to install all windows updates.
See https://github.com/dokan-dev/dokany/issues/187 (2008 r2 is the same as Win7)

Issue #187, opened 2016-03-03
Windows requires a digitally signed driver


qaisjp:
when installing "DokanSetup_redist.exe" in [v1.0.0-RC1](https://github.com/dokan-dev/dokany/releases/tag/v1.0.0-RC1)

![image](https://cloud.githubusercontent.com/assets/923242/13490346/f507382a-e123-11e5-850c-4c3f18aed5b5.png)

I'm on Windows Enterprise 7

Liryna:
Did you update you windows 7 ?
The drivers are signed SHA1/SHA2, you can see it in the file properties.

qaisjp:
![image](https://cloud.githubusercontent.com/assets/923242/13490422/4f57b96c-e124-11e5-8c45-0e7349b46172.png)

I'm on a networked computer so I cannot update this computer. There was a recent update to this computer but I'm not sure if that's what you're talking about

Liryna:
Yes, Thanks this is what I am talking about.

There should be two signature but since your system is not updated it only see the SHA1 that is Microsoft deprecated....since 2016.

qaisjp:
Okay well I've disabled verification using `bcdedit /set nointegritychecks ON`... lets hope it works..

js69:
Is Service Pack 1 installed for that Windows 7? The setup should fail if it isn't but let's verify anyway.

SHA2 for Windows 7 might need this Windows update:
https://technet.microsoft.com/en-us/library/security/3033929

qaisjp:
No idea if SP1 is installed ... like I said I can't install any updates because updates are managed by my sysadmin.

js69:
You could post the setup log files, in your %TEMP% there should be Dokan* text files.

qaisjp:
ok

qaisjp:
here are all the logs
[Dokan_Library_1.0.0_Bundle_20160303092855_000_Dokan_x64.msi.log.txt](https://github.com/dokan-dev/dokany/files/156501/Dokan_Library_1.0.0_Bundle_20160303092855_000_Dokan_x64.msi.log.txt)
[Dokan_Library_1.0.0_Bundle_20160303093340.log.txt](https://github.com/dokan-dev/dokany/files/156502/Dokan_Library_1.0.0_Bundle_20160303093340.log.txt)
[...]

PR #260, opened 2016-06-13
Installer: Make sure that all updates are installed on Win7

Rondom:
On Windows 7 and 2008 R2 the installer did not check whether SP1 and KB3033929 are installed which led to confused users like me or #187 or #244. For details see the individual commits.

I will need to do some testing, to test the different combinations. ETA end of this week!

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,12 @@ You can obtain source files from https://dokan-dev.github.io
 ## Environment
 Dokan works on
  * Windows 10
- * Windows 8.1
  * Windows Server 2012 R2
- * Windows 8
+ * Windows 8.1
  * Windows Server 2012
- * Windows 7
+ * Windows 8
+ * Windows Server 2008 R2 SP1
+ * Windows 7 SP1
 
 ## How it works
 Dokan library contains a user mode DLL (dokan1.dll) and a kernel mode
--- a/dokan_wix/Bootstrapper/Bundle.wxs
+++ b/dokan_wix/Bootstrapper/Bundle.wxs
@@ -59,6 +59,7 @@
     <util:RegistrySearch Id="vcredist_2015_x64_versioncheck" Root="HKLM" Key="SOFTWARE\Classes\Installer\Dependencies\Microsoft.VS.VC_RuntimeMinimumVSU_amd64,v14" Value="Version" Variable="vcredist_2015_x64_version" Result="exists" Win64="yes"/>
     <util:RegistrySearch Id="vcredist_2015_x64_versionnumbercheck" Root="HKLM" Key="SOFTWARE\Classes\Installer\Dependencies\Microsoft.VS.VC_RuntimeMinimumVSU_amd64,v14" Value="Version" Variable="vcredist_2015_x64_versionnumber" Result="value" Win64="yes"/>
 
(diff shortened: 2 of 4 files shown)


Liryna:
:+1: Tell me when I can test it !

Rondom:
@Liryna You can test as soon and as much as you want :-) It works on my patched Win 2008 R2. 

What should be tested (probably in that order):
* without .NET4 (bundle)
* with .NET but without SP1 (both MSI and bundle)
* with .NET and SP1 but without KB3033929 (both MSI and bundle)
* with .NET, SP1 and KB3033929 (both MSI and bundle)

I do not really feel like testing both 32 and 64 bit.

Liryna:
Worked for me :+1: 

Thank you @Rondom for your contributions as always :heart: !!!

Well found for `Wintrust.dll`  check :D

Rondom:
Did some tests and it does not work correctly.  Here is why:

> **Important:** When doing a locale-neutral search for a file, **you must set the MinVersion property to one revision number lower than the actual version you want to search for.** In this example, while we want to find file version 6.0.6001.1751, the MinVersion is set to 6.0.6001.1750. This is because of a quirk in how the Windows Installer matches file versions. More information is available in the Windows Installer documentation.

I have no words for this...

Liryna:
Oh :cry: 
Do we need to reduce the MinVersion so ?

Rondom:
Probably... I really wonder why WiX does not abstract this oddity away...
Also I am seriously interested why it has to be *lower* and not *higher* by one...

Here are my results so far. Tested with DokanSetup-1.0.0.3616.exe (Git bbbdbcb8)

    +----------+-------+-----+-----------+-----------------+---------------+
    |OS        | .NET4 | SP1 | KB3033929 | Bundle-Msg      | MSI-Msg       |
    |----------+-------+-----+-----------+-----------------+---------------|
    | Win7 x86 |   o   |  o  |    o      | No SP1          | No .NET4      |
    | Win7 x86 |   x   |  o  |    o      | No SP1          | No SP1        |
    | Win7 x86 |   x   |  x  |    o      | No KB3033929 ¹  | No KB3033929  |
    | Win7 x86 |   o   |  x  |    o      | No KB3033929 ¹  | No .NET 4.0   |
    | Win7 x86 |   x   |  x  |   (x)²    | No KB3033929 ¹  | No KB3033929  |
    | Win7 x86 |   x   |  x  |    x      |                 |               |
    +----------+-------+-----+-----------+-----------------+---------------+

    ¹ installs bundled software (.NET 4.0 and runtimes) before failing with missing KB3033929
    ² No reboot

None of the error-messages are wrong, but not everything is optimal and one could make the user experience nicer by giving more informative messages by reordering the checks and/or by adding a check for KB3033929 to the bundle (see footnote). I honestly lack the motivation and time for this.  My main goal was that the installer or Dokan does not fail in mysterious ways
As @js69 very nicely put it, supporting such configurations is a valiant but somewhat quixotic effort. **So if someone is valiant and wants to improve the user experience and do the required testing (see above), please go ahead ;-)**

**What I will do is fix the KB3033929-check by submitting a PR and test on Monday or Tuesday.**

PR #266, opened 2016-06-20
dokan_wix: Fix KB3033929-check

Rondom:
**Not yet tested, don't merge**

Due to some oddity, we need to specify a version that is one version
lower than the one we are actually searching for.
References: #260, #244, #187

Quoting WiX-documentation:
Important: When doing a locale-neutral search for a file, you must set
the MinVersion property to one revision number lower than the actual
version you want to search for.


Rondom:
Done tests. Works and detects both missing and installed update correctly.

Liryna:
Ah @Rondom ! Thank you for your time and contribution as always :heart: 
very appreciate !

Issue #249, opened 2016-05-27
MS Word 2007 & 2010 can not save file normally in mirror.exe 


superhq:
# Environment
 * Windows version: Windows Server 2008 R2, Windows 8.1 
 * Processor architecture: X64 
 * Dokany version: 1.0.0-RC3
 * Library type (Dokany/FUSE): Dokany
 * MS Word Version: MS Office 2007 & 2010 
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description
Do the fallowing steps in mirror sample can reproduce the problem:

1. Create a .docx file  (OK)
2. Open the file, write some characters, and save it  (OK)
3. Close the file  (OK)
4. **Open the file again** (OK)
5. Modify it (OK)
6. Save it (ERROR)
 
# Logs
[log.txt](https://github.com/dokan-dev/dokany/files/285872/log.txt)

marinkobabic:
It's strange to me that you at some point get the following output:

###GetFileInfo 1584
GetFileInfo : c:\test\~WRL0002.tmp
    invalid handle

    GetFileInformationByHandle success, file size = 13063

    result =  0
    FileStandardInformation
    DispatchQueryInformation result =  0

Liryna:
The system seems to make a request to GetFileInfo without calling `CreateFile`.
In this case the mirror is calling `CreateFile`
https://github.com/dokan-dev/dokany/blob/master/samples/dokan_mirror/mirror.c#L629

@superhq , can you try to add `FILE_SHARE_WRITE  FILE_SHARE_DELETE` to this `CreateFile` ?

superhq:
@Liryna I try to use CreateFile with share mode FILE_SHARE_READ|FILE_SHARE_WRITE|FILE_SHARE_DELETE in MirrorGetFileInformation function, but result dose not change.
I set a breakpoint in
 [https://github.com/dokan-dev/dokany/blob/master/samples/dokan_mirror/mirror.c#L632](https://github.com/dokan-dev/dokany/blob/master/samples/dokan_mirror/mirror.c#L632)
[...]

PR #273, opened 2016-07-02
SL_OPEN_TARGET_DIRECTORY directory and fcb leak

bailey27:
These changes are primarily for changing the behavior of creating a file when SL_OPEN_TARGET_DIRECTORY.

This is to fix this bug https://github.com/dokan-dev/dokany/issues/249

Now when SL_OPEN_TARGET_DIRECTORY is specified, the driver creates the fcb for the parent dir of the file (instead of for the file itself).  Then it passes the original filename to the dll.  The dll then checks to see if the file exits.  Then the dll opens the parent dir o fthe file.  If the file did not exist but the dir opened successfully, then it sets info to FILE_DOES_NOT_EXIST and returns STATUS_SUCCESS.  Word seems to rely on that behavior.  

The changes also fix an fcb and ccb leak and a memory leak of eventContext when an oplock operation fails.  FsRtlOplockFsctrl() was the one observed failing.  Now when DokanDispatchCreate() does _leave early because of an error, it frees the allocated resources that it was leaking before.

Also, some more debug prints were added.

The changes about SL_OPEN_TARGET_DIRECTORY seem to fix the Word problem.  The Word problem was also caused by the fcb leak.  So even though the fix for the fcb leak wasn't needed to fix this bug with Word, it should be there because a leaked fcb can cause all sorts of hard to debug problems.

### DIFF
--- a/dokan/create.c
+++ b/dokan/create.c
@@ -101,6 +101,9 @@ VOID DispatchCreate(HANDLE Handle, // This handle is not for a file. It is for
   DWORD options;
   DOKAN_IO_SECURITY_CONTEXT ioSecurityContext;
   WCHAR *fileName;
+  BOOL childExisted = TRUE;
+  WCHAR *origFileName = NULL;
+  DWORD origOptions;
 
   fileName = (WCHAR *)((char *)&EventContext->Operation.Create +
                        EventContext->Operation.Create.FileNameOffset);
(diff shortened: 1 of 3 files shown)


Liryna:
Hi @bailey27 ,

Thank you for your contribution and your time on the issue ! very appreciate ! :+1: 

Fix office 2010 and handle correctly SL_OPEN_TARGET_DIRECTORY.

Liryna:
@bailey27 , there is two issues reported by coverity:
fileName is free two times when device is read only: 
https://github.com/dokan-dev/dokany/blob/1bf0c58e8e52bc65334898f152429c298b313fe1/sys/create.c#L649
https://github.com/dokan-dev/dokany/blob/1bf0c58e8e52bc65334898f152429c298b313fe1/sys/create.c#L1116

Also he say that parentDir leak and is never free
https://github.com/dokan-dev/dokany/blob/1bf0c58e8e52bc65334898f152429c298b313fe1/sys/create.c#L1113

bailey27:
Thanks for telling me.  I will go fix it.

bailey27:
I did another pull request.

PR #274, opened 2016-07-02
fix coverity issues with freeing fileName and parentDir when device is…

bailey27:
I moved the getting of the parentDir to after the read-only check to fix the coverity issues it was causing.  This code should have been where it is now, which is right before DokanGetFCB() is called.

### DIFF
--- a/sys/create.c
+++ b/sys/create.c
@@ -626,14 +626,6 @@ Return Value:
                     fileObject->FileName.Length);
     }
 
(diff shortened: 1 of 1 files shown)


bailey27:
I realized just realized now I don't like how it looked (cosmetic). I will fix it and do another pull.

bailey27:
It's there now.

PR #275, opened 2016-07-02
fix coverity issues with freeing fileName and parentDir when device is read-only

bailey27:
I moved the code down to fix the coverity issue with fileName and parentDir if the device is read-only.

Then I realized I should just merge it with the block below it.  Otherwise there is a redundant check for SL_OPEN_TARGET_DIRECTORY.

### DIFF
--- a/sys/create.c
+++ b/sys/create.c
@@ -642,16 +642,13 @@ Return Value:
         __leave;
       }
     }
-    
+
     if (irpSp->Flags & SL_OPEN_TARGET_DIRECTORY) {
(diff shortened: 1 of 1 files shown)


Liryna:
:+1: thank you @bailey27 for the quick fix !

bailey27:
There's yet another problem.  As it is, you can't rename files in the root directory.  I'm fixing that now.

All I can say for myself about that is I did run winfstest on m:\, and it did not complain.

PR #276, opened 2016-07-02
Fix rename in root dir

bailey27:
This has to do with SL_OPEN_TARGET_DIRECTORY

Both the the code I added in the driver, and the code that was already there in dokan\create.c for getting the parent directory would turn a path like \foo.txt into an empty string when trying to get the parent directory.

This was preventing files in the root of the mirror fs from being renamed.

I fixed both of them to get \ in that case.

Also, I noticed that dokan\creat.c was returning STATUS_SUCCESS even if opening the parent dir failed.  It shouldn't do that.  It should return success if the parent can be opened but the child does not exist.  It shouldn't return success if the parent cannot be opened.

### DIFF
--- a/dokan/create.c
+++ b/dokan/create.c
@@ -179,6 +179,11 @@ VOID DispatchCreate(HANDLE Handle, // This handle is not for a file. It is for
     if (lastP) {
       *lastP = 0;
     }
+    
+    if (!fileName[0]) {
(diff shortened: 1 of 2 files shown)


Liryna:
Thank you for the qucik fix @bailey27  ! :heart:

PR #277, opened 2016-07-03
should not free eventContext, ccb, or fcb if oplock is pending

bailey27:
I noticed this happening 

```
M:\ echo blah > foo.txt
M:\ del foo.txt
[hang]
```

The cmd window is hung.  The driver and mirror are still ok. 

If I disabled Windows Defender real-time scanning, it didn't happen.

When I fixed the leaks that happened because of leaving early from the failed oplock operation, I also "fixed" a non-leak in the case above it, where the the oplock is pending. In that case, the operation is registered as pending, and no additional freeing is necessary or desirable.

Here is the fix


Liryna:
Hi @bailey27 ,

Oh yes, I did not seen :( you are right, this has to be keeped !
Thank you ! :+1:

PR #280, opened 2016-07-05
sys: DokanCompleteQueryVolumeInformation check for NULLs

taruti:
Does not solve all issues but seems to help a bit.


taruti:
Even with https://github.com/dokan-dev/dokany/pull/282 in I can hit this with the CCB being NULL.

Issue #272, opened 2016-07-01
BSODs while using dokanfuse.dll with fuse-jna


markuskreusch:
# Environment
 * Windows version: Windows 7 Home Premium / 6.1 (Build 7601: Service Pack 1)
 * Processor architecture: x64
 * Dokany version: 1.0.0-RC3
```
>dokanctl.exe /v
Driver path: 'C:\Windows\system32\drivers\dokan1.sys'
dokanctl : May 11 2016 13:51:28
Dokan version : 100
Dokan driver version : 0x190
```
 * Library type (Dokany/FUSE): dokanfuse
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)

    similar to #230 but does not involve fltmgr.sys
  - [ ] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
    see description
  - [ ] I can always reproduce the issue with the provided description below.    
    Not always but the problem occured three times yesterday and the day before after running the file system for at most 20 minutes each time.
  - [x] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description

I mounted a file system implemented in java using fuse-jna (after small adaptions to make fuse-jna work with dokanfuse / windows). I get BSODs when using dokanfuse, sometimes immediately after mounting and sometimes after using the filesystem for some time. Dump analysis showed it is somehow related to dokany: dokan1!IsMountPointDriveLetter.

I am not 100% sure if this is unspecific to my user-mode driver implementation. My java code should be fine (uncommon to produce such problems) but maybe fuse-jna has some problems. After testing with the fuse mirror sample I had no problems but this might only be because I used my implementation much more than the sample. My experience in analysing dumps is limited so I hope you can look into this and give me some pointers. If I should run further analysis on my system I can do so. In addition I will try to reproduce the problem using the fuse mirror sample during the next days.

# Logs

Dumps of all three BSODs: [dumps.zip](https://github.com/dokan-dev/dokany/files/343048/dumps.zip)

!analyse -v of one of them:

```
*******************************************************************************
*                                                                             *
*                        Bugcheck Analysis                                    *
*                                                                             *
*******************************************************************************

UNEXPECTED_KERNEL_MODE_TRAP (7f)
This means a trap occurred in kernel mode, and it's a trap of a kind
[...]

Issue #281, opened 2016-07-06
The directory name is invalid 


cpptinybird:
# Environment
 * Windows version:  Windows7 sp1
 * Processor architecture: x86
 * Dokany version: 0.8
 * Library type (Dokany/FUSE):  Dokany
 
# Check List
  - [ ] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [ ] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [ ] I can always reproduce the issue with the provided description below.
  - [ ] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description
I want to Implement a file system like this (Using "mirror" sample): When double-click any file in my dokan device,it will open the specific file(like "C:\1.txt") for me.So I changed MirrorCreateFile :
```
static NTSTATUS DOKAN_CALLBACK MirrorCreateFile() 
{
    WCHAR filePath[MAX_PATH] = L"C:\\1.txt";
    ...
    //GetFilePath(filePath, MAX_PATH, FileName); 
}
```

But when I open text file (double-click),Windows will popup a window like this:
![Alt text](http://img.blog.csdn.net/20160706145952059)
(It means "The directory name is invalid ")

Now I want to know ,why Windows thinks my text file(unzlibcmpress.py) is a directoty?

# Logs
Please attach in separate files: mirror output, library logs and kernel [logs](https://github.com/dokan-dev/dokany/wiki/How-to-Debug-Dokan).
In case of BSOD, please attach minidump or dump analyze output.

Liryna:
Hi @cpptinybird ,

I don't understand the issue, can you be more specific ?

cpptinybird:
@Liryna I'm Sorry ,My English is poor...
The only code I changed was here ,you can see it in this picture:
![Alt text](http://img.blog.csdn.net/20160706151930026)
Then I mounted directory ("test_dokan") ,But when I open a file ,this happened:
![Alt text](http://img.blog.csdn.net/20160706151938994)
Why Windows does not open "C:\1.txt" for me and thinks this file is a  directory ?

cpptinybird:
The error on Windows MessageBox means "The directory name is invalid ".

Liryna:
@cpptinybird 
[...]

Issue #230, opened 2016-05-10
BSOD with lots of concurrent mount+unmounts 


taruti:
# Environment
 * Windows version: Windows 7
 * Processor architecture: x64 and x86
 * Dokany version: master 249d4f5 and 1.0RC2
 * Library type (Dokany/FUSE): kernel driver
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [ ] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [ ] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description

I am looking into a BSOD that comes from running lots of concurrent mounts + unmounts, but still trying to isolate it and create a minimal example.

Thus far:
* Mounts + unmounts from multiple threads help for getting the BSOD
* Typially I get it after a couple minutes of a stress testing
* Remote kernel debugging makes the BSOD go away (perhaps due to slowness and thus not hitting the concurrency issue) and not occur in hours of stress testing
* This is probably related to insufficient locking somewhere in the kernel driver
* Working on isolating a tiny program to recreate this

Not sure how an issue like this could be reproducible by the mirror sample.

# Logs

Here is !analyze -v from a half dozen kernel dumps.

```

Microsoft (R) Windows Debugger Version 6.3.9600.17336 AMD64
Copyright (c) Microsoft Corporation. All rights reserved.


Loading Dump File [C:\k\MEMORY1.DMP]
Kernel Summary Dump File: Only kernel address space is available

WARNING: Whitespace at end of path element
Error: Empty Path.
Symbol search path is: srv*C:\symbols\ms*http://msdl.microsoft.com/download/symbols

Executable search path is: 
Windows 7 Kernel Version 7601 (Service Pack 1) MP (4 procs) Free x64
Product: WinNt, suite: TerminalServer SingleUserTS Personal
Built by: 7601.19045.amd64fre.win7sp1_gdr.151019-1254
Machine Name:
Kernel base = 0xfffff800`02a53000 PsLoadedModuleList = 0xfffff800`02c9a730
Debug session time: Mon May  9 13:04:35.777 2016 (UTC + 1:00)
System Uptime: 0 days 0:01:06.244
Loading Kernel Symbols
...............................................................
................................................................
................
[...]

PR #282, opened 2016-07-06
sys: Add IsUnmounting to the DCB

taruti:
As suggested by @marinkobabic on https://github.com/dokan-dev/dokany/issues/272

I can get this to hit occasionally in stress tests. 

However this does not replace https://github.com/dokan-dev/dokany/pull/280

### DIFF
--- a/sys/dokan.h
+++ b/sys/dokan.h
@@ -225,6 +225,7 @@ typedef struct _DokanDiskControlBlock {
   USHORT UseMountManager;
   USHORT MountGlobally;
   USHORT FileLockInUserMode;
+  USHORT IsUnmounting;
 
   // to make a unique id for pending IRP
   ULONG SerialNumber;
--- a/sys/event.c
+++ b/sys/event.c
@@ -300,6 +300,7 @@ DokanCompleteIrp(__in PDEVICE_OBJECT DeviceObject, __in PIRP Irp) {
   PLIST_ENTRY thisEntry, nextEntry, listHead;
   PIRP_ENTRY irpEntry;
   PDokanVCB vcb;
+  PDokanDCB dcb;
   PEVENT_INFORMATION eventInfo;
 
   eventInfo = (PEVENT_INFORMATION)Irp->AssociatedIrp.SystemBuffer;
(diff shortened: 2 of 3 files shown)


Liryna:
Hi @taruti ,

Does this improve one of the issue you are facing ?

marinkobabic:
I will do some small improvements and see if you still can reproduce your issue. Just wait for a new pull request.

taruti:
@Liryna With both this and #280 (CCB is NULL) I am seeing less crashes with the stress test in #230.

#280 seems to be encountered more frequently than this, but this is also encountered sometimes.

marinkobabic:
@taruti 
Please can you check if this solves the ccb null problem https://github.com/marinkobabic/dokany/tree/FixNullIReference If we do not understand why is ccb null, we will not be able to fix the issue definitely. But with your help we should be able to identify the source of the problem.

taruti:
With https://github.com/marinkobabic/dokany/tree/FixNullIReference I get a fast crash without good logs.

Tested with example code from #230 and Windows 7.

marinkobabic:
If you put your fix with ccb into FixNullReference code, then you don't have the crash?

taruti:
Still have a crash. I think your patch introduces a new crash. Also with two flags for Mounted/IsUnmounted most of the code checks only one of them.

Does that branch work for you with the example code from #230 on Windows 7?

marinkobabic:
This items can be null, which is fine:

- ccb
- FsContext2

So the question which of this items is null in your case, when you get the crash:

- IrpEntry->FileObject
- vcb
- vcb->Dcb

marinkobabic:
Yes it works fine, The problem is that I'm testing on Windows 10 x64 where you don't have this issue?

taruti:
MS used to have VMs for testing IE that included a Windows 7 image.

Yes, this seems Windows 7 specific.

marinkobabic:
So I will setup Windows 7 x64 :smile: I agree with your changes but I have no idea why FileObject, vcb or dcb should be null. So I must understand it.

taruti:
Both 32 and 64 bit win 7 seems to be enough for a crash with the branch.

marinkobabic:
You can run mirror using the new version without any issues? Only the stress tests causes the driver to fail?

taruti:
In my Windows 7 tests:

Branch | Mirror | Stress test from #230
------------ | ------------- | --------------------
master | ok | crash
#281 | ok | crash less often
#282 | ok | crash less often
FixNullIReference branch | crash fast | crash fast

marinkobabic:
have returned wrong state when the volume is not mounted, please can you try again just the mirror?

taruti:
https://github.com/marinkobabic/dokany/tree/FixNullIReference with the second patch fails to compile:
```uninitialized local variable 'ccb' used``` in volume.c line 197. Is there a commit missing from there?

marinkobabic:
It’s fixed



Von: Taru Karttunen [mailto:notifications@github.com]
Gesendet: Mittwoch, 6. Juli 2016 15:55
An: dokan-dev/dokany <dokany@noreply.github.com>
Cc: Babic Marinko <marinko.babic@ruf.ch>; Mention <mention@noreply.github.com>
Betreff: Re: [dokan-dev/dokany] sys: Add IsUnmounting to the DCB (#282)



https://github.com/marinkobabic/dokany/tree/FixNullIReference with the second patch fails to compile:
uninitialized local variable 'ccb' used in volume.c line 197. Is there a commit missing from there?

taruti:
After the third commit it compiles but crashes instantly also with mirror.

Issue #287, opened 2016-07-09
event context leak on every file lock operation when FileLockInUserMode is false


bailey27:
# Environment
 * Windows version: Windows 10
 * Processor architecture: x64
 * Dokany version: current source
 * Library type (Dokany/FUSE): Dokany
 
# Check List
  - [x ] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x ] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x ] I can always reproduce the issue with the provided description below.
  - x[ ] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description

It looks to me that DokanDispatchLock() normally leaks an event context every time it is called when FileLockInUserMode is false (which is the default setting).

This happens because it always allocates an eventContext, but when FileLockInUserMode is false, the operation shouldn't be and is not normally sent to user mode, and the irp is completed by FsRtlProcessFileLock() and not the dokany driver.  So the event context is never freed.

The only time it doesn't leak when FileLockInUserMode is false is if FsRtlCheckOplock() returns pending. But if FsRtlCheckOplock() does return pending, then the lock request will be sent to user mode, which isn't supposed to happen when FileLockInUserMode is false.

I think what it should do is, if FileLockInUserMode is false, then not allocate an event context.  And it should pass a different (new) callback to FsRtlCheckOplock() that places the operation on work queue instead of using DokanOplockComplete which will send the operation to user mode.  That is what the oplock complete in fastfat does.

I am not sure how to cause an oplock break on a lock operation to happen though.  So I don't know how to test that.

In general, I'm not having much luck at writing a working program that opens a file and breaks the oplock held by another process on some other operation besides create (e.g. a write).  But I'm working on that.

But I think the leak could be serious if somebody ran an app that did lots of file locking.

To see the leaks happening, I complied the second example on this page

https://msdn.microsoft.com/en-us/library/windows/desktop/aa365204(v=vs.85).aspx

into locktest.exe

I added this code to the driver for tracking allocates and frees of event contexts.

http://pastebin.com/VFmiv8x0

I did some normal things like dir, and cat and dd'd 1MB to the mirror drive to make sure my allloc counting was working.

Then I ran locktest.exe foo.dat

I can see towards the end of my log file that the event contexts are being leaked on every lock operation.

# Logs

[...]

PR #290, opened 2016-07-18
FileLockInUserMode kernel memory leak fix

bailey27:
This is for https://github.com/dokan-dev/dokany/issues/287


Liryna:
Thank you @bailey27 for your contribution ! :heart: 
And well done for seeing it again :+1: 
(I made a patch and pushed it directly because of the git history of your branch)

Issue #294, opened 2016-07-21
Filename comparisons when finding fcb are always case-sensitive


bailey27:
# Environment
 * Windows version:  Windows 10
 * Processor architecture: x64
 * Dokany version: current source
 * Library type (Dokany/FUSE):  Dokany
 
# Check List
  - [x ] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x ] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x ] I can always reproduce the issue with the provided description below.
  - [ x] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description

The issue is that, in the kernel-mode driver, when DokanGetFCB() goes to find an existing fcb for a file, it does the string comparison in a case-sensitive fashion.

I think this generally isn't a problem, but, because NTFS is (normally) case-insensitive, you could end up with two FCBs for the same file, which isn't how it's supposed to be.

I was able to get it to happen
```
[DokanFS]   Create: ProcessId 612, FileName:\FOO.txt
[DokanFS]   DokanGetFCB has entry FileName: \ FileCount: 3. Looking for \FOO.txt
[DokanFS]   DokanGetFCB has entry FileName: \foo.txt FileCount: 1. Looking for \FOO.txt
[DokanFS]   Allocate FCB for \FOO.txt
[DokanFS]   Create: FileName:\FOO.txt, status = 0x00000103
```

As you can see, even though there is already an fcb for \\foo.txt, it is allocating a new one for \\FOO.txt.

I changed the code to use RtlEqualUnicodeString() and also use the SL_CASE_SENSITIVE in irpSp->Flags to determine if the comparison should be case-sensitive.

Here is my branch
https://github.com/bailey27/dokany/commit/5960d59a1fa0d1f09d61bc627e8a805cd10c8586

Now it finds the fcb for \\foo.txt when it goes to open \\FOO.txt

```
[DokanFS]   Create: ProcessId 4736, FileName:FOO.txt
[DokanFS]   DokanGetFCB has entry FileName: \ FileCount: 2. Looking for \FOO.txt
[DokanFS]   DokanGetFCB has entry FileName: \foo.txt FileCount: 1. Looking for \FOO.txt
[DokanFS]   Found existing FCB for \FOO.txt
[...]

PR #297, opened 2016-07-21
Do filename comparisons in DokanGetFCB() either case sensitive or not…

bailey27:
… depending on flag SL_CASE_SENSITIVE

This is about https://github.com/dokan-dev/dokany/issues/294


Liryna:
Perfect pull request ! :+1: 

Thank you @bailey27 !

PR #302, opened 2016-07-27
Fix BSOD mount/unmount, BitDifender, symantec

marinkobabic:
There are a lot of changes. The major change is, when the device has ReferenceCount > 0 it will not be deleted. When ReferenceCount reachs 0 then it will be deleted delayed.

### DIFF
--- a/sys/fscontrol.c
+++ b/sys/fscontrol.c
@@ -539,8 +539,16 @@ NTSTATUS DokanMountVolume(__in PDEVICE_OBJECT DiskDevice, __in PIRP Irp) {
     DDbgPrint("   Not DokanDiskDevice\n");
     return status;
   }
+
+  if (IsDeletePending(dcb->DeviceObject)) {
+      DDbgPrint(" This is a remount try of the device");
+      return STATUS_DEVICE_REMOVED;
+  }
+
   BOOLEAN isNetworkFileSystem =
       (dcb->VolumeDeviceType == FILE_DEVICE_NETWORK_FILE_SYSTEM);
+
+  DDbgPrint(" Mounting volume using MountPoint %wZ device %wZ \n", dcb->MountPoint, dcb->DiskDeviceName);
 
   if (!isNetworkFileSystem) {
     status = IoCreateDevice(DriverObject,               // DriverObject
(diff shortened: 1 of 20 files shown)


sys/fscontrol.c
```diff
@@ -614,14 +624,18 @@ NTSTATUS DokanMountVolume(__in PDEVICE_OBJECT DiskDevice, __in PIRP Irp) {
     RtlCopyMemory(dokanControl.UNCName, dcb->UNCName->Buffer,
                   dcb->UNCName->Length);
   }
-  mountEntry = FindMountEntry(dcb->Global, &dokanControl);
-  if (mountEntry != NULL) {
+  mountEntry = FindMountEntry(dcb->Global, &dokanControl, TRUE);
+  if (mountEntry != NULL && &mountEntry->MountControl) {
```
Maxhy: What do you attempt to do here? I may be wrong but this check looks invalid to me.

marinkobabic: If the MountControl pointer is still valid. Depending on timing it could be null if at the same time somebody removed the mounEntry.

Afterster: But MountControl is not a pointer only mountEntry is, it can be zero-filled but its address will never be NULL.

sys/init.c
```diff
@@ -368,31 +423,313 @@ InsertMountEntry(PDOKAN_GLOBAL dokanGlobal, PDOKAN_CONTROL DokanControl) {
...
+        deviceEntry = CONTAINING_RECORD(entry, DEVICE_ENTRY, ListEntry);
+        if (deviceEntry) {
+            DDbgPrint("  There is a device for delayed delete. Counter %lu \n", deviceEntry->Counter);
+            InterlockedIncrement((LONG *)&deviceEntry->Counter);
+            
+            BOOLEAN canDeleteDiskDevice = deviceEntry->DiskDeviceObject->ReferenceCount == 0;
+
+            dcb = deviceEntry->DiskDeviceObject->DeviceExtension;
+
+            if (deviceEntry->Counter > 3) {
```
marinkobabic: You mean put some comments on top of the method?

sys/fscontrol.c
Maxhy: But MountControl is not a pointer only mountEntry is, it can be zero-filled but its address will never be NULL.

sys/init.c
Maxhy: I'm more talking about this specific `> 3` check, it's not self explanatory and you have to read a lot of code to get why there is such check.

sys/fscontrol.c
marinkobabic: You are right. The check is invalid and because of the concept change not needed anyway.

Liryna:
Hi @marinkobabic ,

Thank your for your contribution !
I only go an issue when using the mount manager option `/o`. The driver letter can no longer is be assigned and the device is still showed on explorer.

sys/init.c
marinkobabic: Will add some notes. It would no be really delayed if we wouldn't have such value. But you are right

marinkobabic:
Ok. Will check what is there going on.

Liryna:
I have run @taruti https://github.com/dokan-dev/dokany/issues/230#issuecomment-218377179 sample without issue (else than memory leak :smile: ) during 45min without crash.
@taruti did you also try the sample during your test ?

taruti:
Will test on monday and complain if it still exists - many thanks for the fixes.

PR #303, opened 2016-07-29
Fix BSOD mount/unmount, BitDifender, symantec desktop encryption

marinkobabic:
Here you have the updated pull request which also works using the mountmanager.

### DIFF
--- a/dokan/dokan.c
+++ b/dokan/dokan.c
@@ -362,29 +362,31 @@ UINT WINAPI DokanLoop(PDOKAN_INSTANCE DokanInstance) {
   }
   RtlZeroMemory(buffer, sizeof(char) * EVENT_CONTEXT_MAX_SIZE);
 
-  device = CreateFile(GetRawDeviceName(DokanInstance->DeviceName, rawDeviceName,
-                                       MAX_PATH),         // lpFileName
-                      GENERIC_READ | GENERIC_WRITE,       // dwDesiredAccess
-                      FILE_SHARE_READ | FILE_SHARE_WRITE, // dwShareMode
-                      NULL,          // lpSecurityAttributes
-                      OPEN_EXISTING, // dwCreationDistribution
-                      0,             // dwFlagsAndAttributes
-                      NULL           // hTemplateFile
-                      );
-
-  if (device == INVALID_HANDLE_VALUE) {
-    DbgPrint(
-        "Dokan Error: CreateFile failed %ws: %d\n",
-        GetRawDeviceName(DokanInstance->DeviceName, rawDeviceName, MAX_PATH),
-        GetLastError());
-    free(buffer);
-    result = (DWORD)-1;
-    _endthreadex(result);
-    return result;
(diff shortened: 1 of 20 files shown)


Liryna:
Hi @marinkobabic ,

Thank you, the quick fix work correctly !
Mount managers work correctly now :+1: 

Does the lock should not be removed here https://github.com/dokan-dev/dokany/pull/303/files#diff-a3d0440cc7bd9a2cd87e16827fe0dfbdR842 ?
QueryDeviceRelations is no longer called, is there a reason for it ? Should we remove the function ?
https://github.com/dokan-dev/dokany/pull/303/files#diff-6225bb782de1b5ff3af319d3add27aa9R51

I cast the `Flags` to PLONG for fixing warning, is that alright for you ?
https://github.com/dokan-dev/dokany/compare/FixBSOD?expand=1#diff-044fe7c5cc010c659ea4ee15d12dbebeR676

(I failed the merge the patch is currently in FixBSOD branch https://github.com/dokan-dev/dokany/tree/FixBSOD)

Here is a clean diff: https://github.com/dokan-dev/dokany/compare/FixBSOD?expand=1

marinkobabic:
IoReleaseRemoveLock(&dcb->RemoveLock, Irp);

must be done because the processing is finished. The lock is just there so that nobody removes the device while we are processing the request.

QueryDeviceRelations did not return proper information and verifier complained about it. If we fix it to work properly, we can keep the function because of later support of pnp.

The cast should be fine.

Thanks
Marinko

Liryna:
```
  if (IsDeletePending(DeviceObject)) {
    DDbgPrint("Device is deleted, so go out here \n");
	IoReleaseRemoveLock(&dcb->RemoveLock, Irp);
    return STATUS_NO_SUCH_DEVICE;
  }
```
like that ?

I add comment in QueryDeviceRelations therefore :+1: 

Thanks !

marinkobabic:
Hi Lyrina

 

This looks good :)

 

Thanks

Marinko

Liryna:
https://github.com/dokan-dev/dokany/compare/FixBSOD?expand=1#diff-0d281a5a20ad484e35eb5d133d0af346R369
`CreateFile` has been moved to the while but it still return in case of failure https://github.com/dokan-dev/dokany/compare/FixBSOD?expand=1#diff-0d281a5a20ad484e35eb5d133d0af346R387
is this made in purpose ?

marinkobabic:
Hi Lyrina

This is correct. The problem before was, that we had a handle to a removed device, which often caused a BSOD. Here we close the handle and create a new one. If the device is not there, we will have an invalid handle and we will go out.

Thanks
Marinko

Liryna:
Thank you @marinkobabic for making this clear and for the valuable contribution :+1: ! 
I will release a RC4 today of it

Liryna:
Hi @marinkobabic ,

Can you contact me by mail at liryna.stark@gmail.com ?

Issue #210, opened 2016-04-04
Rewrite user-mode threading to work with overlapped IO/IOCP


Corillian:
The current threading model used by Dokan is a major performance bottleneck in any user-mode driver that can make asynchronous IO calls - for example a driver that provides access to an FTP server. The OS can potentially make thousands of IO calls a second but with a limit of 15 threads and no async callback mechanism those threads will spend most of their time blocking on IO operations (in the FTP example they will block on socket read/writes).

To solve this I propose the following:

1. `DeviceIoControl()` in dokan.c needs to use an IO completion port for dispatching IO events to the user-mode driver. IO events will automatically be handled on the system thread pool and neither Dokan nor the user-mode driver are required to manage their own threads nor worry about the number of threads that are allocated.
2. A flag should be added to Dokan context information provided to the event handler that allows it to specify whether or not the event will be handled asynchronously. This is because the end of an asynchronous call requires special handling - it must explicitly tell Dokan to send a response to the kernel.
3. The Dokan user-mode `Dispatch*()` functions will need to be split in two: one half (`BeginDispatch*()`) for event setup and the call into the driver event handler and one half (`EndDispatch*()`) for validation, cleanup, and sending the result to the kernel. Finalization for an asynchronous event (`EndDispatch*()`) can be initiated by a `DokanFinishAsyncEvent()` function called from the driver that takes a pointer to the context information as an argument.
4. Currently information is passed to the user-mode driver and back to the kernel via function arguments. This would need to be changed to a pointer to a structure that's internally held by the Dokan event context info so that `DokanFinishAsyncEvent()` knows where to find the information it needs to pass back to the kernel.
5. Dokan context objects should be pooled and push/popped from a global stack to prevent the driver from constantly hitting the memory allocator. A robust implementation should also include time stamps so that a number of context objects in the global pool beyond some threshold can be cleaned up after a some amount of time passes to account for large temporary spikes in activity.

Unfortunately this both a non-trivial undertaking and would require breaking API changes however my driver is already running into problems with throughput due to restrictions imposed by the current Dokan threading model.

Liryna:
Great propsition @Corillian ! I totally agree with you :heart: 

Removing the threads limitation would enhance a lot the dokan speed also !

If I understood well, most of the changes are going to be in the Dokan Library and we are keeping the `DeviceIoControl` for communication Library/Kernel, are you we gonna break API compatibility in this communication ?

[...]

Issue #197, opened 2016-03-16
Cannot mount VHD


Casimir666:
Hi,
I have noticed this issue with the mirror sample application : 
- Create a VHD using Windows "Computer management" (manage / Disk management / Create VHD)
- Start the mirror sample

Attach the VHD using "Computer management" will then display an error "The device does not recognize the command."

My environment is :
- Windows Server 20012 R2 (x64)
- Dokan version : dokany-1.0.0-RC2
- Mirror log file : [MountVHD.zip](https://github.com/dokan-dev/dokany/files/173635/MountVHD.zip)

This problem is reproductible on other OS such as Windows 7, but the error message can be slightly different.

Liryna:
Hi @Casimir666 ,

If I understood well, you are asking Windows to load a VHD device that is inside the mirror device ?

it seems that you have a Lock / UnLock issue.
```
###Lock 0007
LockFile c:\Temp\1-Empty.vhd
	success

###Lock 0007
UnlockFile c:\Temp\1-Empty.vhd
	success

###Lock 0007
LockFile c:\Temp\1-Empty.vhd
	success

###Lock 0007
UnlockFile c:\Temp\1-Empty.vhd
	success

###Lock 0007
LockFile c:\Temp\1-Empty.vhd
	failed(33)
Create got unknown error code 33
###Lock 0007
LockFile c:\Temp\1-Empty.vhd
	failed(33)
Create got unknown error code 33
```

Can you try to run the mirror with one thread `/t 1` ?

Otherwise try to change mirror code Lock/Unlock to always return Success to see if it improve the situation.

Casimir666:
Hi @Liryna,
Thanks you for your answer.

You perfectly understood, I'm trying to mount VHD inside mirror device.

I tried the the /t 1 option, but the problem is the same :
[MountVHD_onethread.zip](https://github.com/dokan-dev/dokany/files/175812/MountVHD_onethread.zip)

I have noticed exactly the same error message if i try to boot an Hyper-V machine on a mirrored VHD. Do you have any hints or tests I can do?

Liryna:
I have changed mirror code on my side to always return success for lock/unlock.
I have helped...until I get a bsod.

This can be reproduced easily.

```
kd> !analyze -v
*******************************************************************************
*                                                                             *
*                        Bugcheck Analysis                                    *
*                                                                             *
*******************************************************************************

PNP_DETECTED_FATAL_ERROR (ca)
[...]

PR #307, opened 2016-08-03
Implemented asynchronous IO #210

Corillian:
This PR was a tremendous amount of work and more or less consists of a full rewrite of dokan.dll. It will need a _significant_ amount of testing before it's ready to be merged into master. We really need to add a large number of unit tests. Notable issues:

1. Significant changes to the Dokan API
2. `FILE_OPEN_FOR_BACKUP_INTENT` in create.c has been removed for now so that it can be more thoroughly investigated
3. Initialization is been moved outside of `DLL_PROCESS_ATTACH` and `DLL_PROCESS_DETACH` to `DokanInit()` and `DokanShutdown()`. This solves the problem of missing DLL's in `DLL_PROCESS_DETACH` and more importantly allows the user to hook into Dokan memory allocations.

Lastly I still need to profile the driver for an optimization pass.


Corillian:
Also FUSE hasn't been tested at all.

Liryna:
Thank you @Corillian for the pull request! 

About unit test there is already some running on appveyor. A big par of the test seems to fail, can you take a look at them?

Corillian:
Compiling FUSE under certain circumstances seems to have failed because the Windows thread pool headers couldn't be located. I don't see any unit tests listed anywhere in the appveyor - only building.

Liryna:
@Corillian https://ci.appveyor.com/project/Maxhy/dokany/build/1.0.0-683/job/hretil424vy6f45t

At the bottom you will have the output of winfstest and fsx.
https://github.com/Liryna/winfstest?files=1

Corillian:
Ah ok, cool, thanks! I'll take a look on Monday.

Corillian:
Ok nearly all of the unit tests are passing. In WinFSTest I get a couple of failures for the regular Windows filesystem however it looks like I have 2 extra failures. FSX seems to be intermittently failing on appveyor, I can only assume it has something to do with async IO however I haven't been able to reproduce the failures on my system yet.

I'll hopefully finish up these last couple of issues next week Monday.

Liryna:
:+1: This begin to looks very good !
Can't wait to review it ^^ !

Liryna:
@Corillian Just wanted to say that I seen you are creating a project based on some WinFSTest .
I think this is a good idea and it will be more easy to add test from issue reported by users !

I will try to give you a hand on it and move all test to the project you created (if you think it is also a good idea).
This is in the long time TODO !

Corillian:
So I am in the middle of Windows security hell. SetFileSecurity keeps failing with ERROR_INVALID_OWNER and I'm not entirely sure why. Documentation for Windows security kind of sucks and I've tried about 1000 different things so I'm starting to run out of ideas. Aside from that I think there's a couple of race conditions in Mirror I need to lock down but everything else is ready to rock and roll. I've also added code that detects FILE_FLAG_BACKUP_SEMANTICS. If I'm not able to get SetFileSecurity sorted out soon I'll probably just commit what I have and see if anybody can think of what could be wrong.

Corillian:
Also the extra weird thing is my code for SetFileSecurity works just fine with my own unit tests. It only fails in the python winfstest unit tests. I'm wondering if it's because it creates a child process and that somehow changes how security is checked. Unfortunately winfstest works fine w/ the regular filesystem so obviously something isn't right in the driver.

Corillian:
For reference to myself as much as anybody else one thing I haven't tried yet is `Get/SetKernelObjectSecurity()`. Documentation is available here: https://msdn.microsoft.com/en-us/library/windows/desktop/aa364399(v=vs.85).aspx

Corillian:
Nevermind you're supposed to use `SetSecurityInfo or SetNamedSecurityInfo` instead of `SetKernelObjectSecurity()` which is what I've been using =(.

Liryna:
Does your changes of `FILE_FLAG_BACKUP_SEMANTICS` flag handle the case describe here https://github.com/dokan-dev/dokany/pull/273 ?
FILE_DOES_NOT_EXIST and returns STATUS_SUCCESS when file does not exist but folder can be opened.

Are you still facing issue with filesecurity or it has been fixed with https://github.com/dokan-dev/dokany/pull/307#issuecomment-241935238 ?

Corillian:
Alright I've done my final commit until you guys have a chance to do a review. Some of the unit tests concerning `SetFileSecurity` are still failing and I have no idea why. I haven't worked with Windows security all that much, the documentation for it sucks, so hopefully somebody who's more familiar with its nuances can spot what's wrong because at this point I'm running out of time.

Everything else seems to be working fine though it's possible there are some race conditions to to a lack of clarity over what which Windows file system functions need to be manually synchronized. Honestly I think we should deprecate Mirror as an example and replace it with an in-memory filesystem that properly emulates all of the functionality. This would save an incredible amount of time trying to figure out how kernel stuff maps back into user space.

Also the new doxygen documentation is a bit broken with this branch since I renamed some of the callback functions to remove conflicts with Windows macros and I also haven't bothered to write the documentation for the new async IO stuff yet. I'll take care of that after the code review is complete.

Liryna:
Thank @Corillian for all this work ! :+1: 

Right after dokan 1.0.0 release that should come during the next week. I will be happy to review it !

Did you test it with heavy software using the dokan device ? (Visual studio, Office,...)
(FSX seems to still have some issues on appveyor)

Corillian:
It looks like there's some sort of race condition that FSX exposes. My own local runs of FSX always succeed but I've been running debug versions so maybe I need to try release. I haven't tried Office or Visual Studio. If I have time next week I'll do a bit more testing.

magol:
I have not reviewed the details, but this looks to be a substantial improvement of Dokan.

But I do not agree that this is appropriate to be released in version 1.1.0 of Dokan. This is a big breaking changes in the API, and according semver, it should be released in a major version.

As much as I want this to be released as soon as possible, we should wait to version 2.0.0. On the other hand, no one is saying that there must be a long time between version 1.0.0 and 2.0.0 :-)

Liryna:
This is a big change I agree but as explained here https://github.com/dokan-dev/dokany/wiki/How-to-package-your-application-with-Dokan#dokan-versioning
We us Major api for kernel<->library compatibility break.

If we move to 2.0.0, all applications will need to install a new version of dokan when they could technically use a possible 1.0.0 driver already installed.

But we will see how this is going to happen. I will first create a develop branch for next features and keep the master for hot fix.

@Rondom Your pull request is the next in my TODO :+1: I just don't know when will have the time for it

Rondom:
I have not looked at the code at all, so my question might sound stupid: How hard would it be to provide a wrapper for the old API? Once the kernel-interface changes and there is a 2.0.0, we could deprecate or remove the old API.

magol:
@Liryna Then I understand how you think, there are always several ways of looking at things. :-)
It gets a little strange for the users of the API, but it is a disadvantage we have to accept. Your motivation is clearly much stronger.

Corillian:
Hey @Liryna now that 1.0 has been released do you have any idea of when you'll get to the review of this PR? I think the longer we wait the harder it's going to be to merge and fix the couple of outstanding issues with Mirror.

Liryna:
Hey @Corillian !
It is in my todo ! Just has to release a 1.0.1 with the main fix reported (in the next week normally) and I move to your PR ❤️ !

magol:
Is there any idea to already start to experimenting with a C#  implementation of this?

Liryna:
Not for now 😃 I will probably need your help on this @magol !

Liryna:
Hi @Corillian ,

I started to test your PR. I see and feel that it is much faster that’s for sure :+1: 
But I am afraid that there is some race conditions :cry: . I get different errors on software I tested .
Creating and mounting VHD does not work.

I often get the error `Insufficient system resources exist to complete the requested service.` in difference conditions. Like copy lots of files or try to build a project with VS on the mirror.
I guess this happens when the mirror has issues to allocate thread or such ?

Corillian:
Thanks for testing it out @Liryna, I have a couple of questions:

1. Which C++ project are you trying to compile? I would like to try compiling it myself to replicate your results.
2. Which VHD are you trying to mount and how are you trying to mount it?
3. Where are the insufficient resource errors codes being returned? I'm assuming they're coming from the user-mode dokan.dll, are they being generated by calls to `DeviceIoControl()`?
4. Did you try disabling `USE_ASYNC_IO` to see if it made any difference? It's defined at the top of mirror.c.
5. I don't suppose you gained any additional insight into the best way to handle the outstanding issues with properly setting file security?

Liryna:
@Corillian 

 1) It is a newly created VS C++ project (not empty with default option that generate a simple main return 0).

 2) VHD is created from `disk management` (Action -> Create VHD). It create a wrong parameter error. It is probably for the same reason that https://github.com/dokan-dev/dokany/issues/197 Forget about it for now :smiley: 

 3) It seems like it is an error of DeviceIoControl as you said (The c++ VS project build):
```
Dokan Information: SendEventInformation() with NTSTATUS 0x0, context 0xa9ef6130, and result object 0x0000022CA9ED6C20
Dokan Error: Dokan device result ioctl failed for wait with code 998.
```
 Full logs of the build: [test.txt](https://github.com/dokan-dev/dokany/files/507036/test.txt)

4) Same error with `USE_ASYNC_IO 0` :cry: 

5) Not for now, I have only time for now to help testing sorry :cry: but I will look into it !

Also, did you rewrite the mount/unmount process ? this seems to be very fast !

Corillian:
Thanks for the info, I don't see any mention of insufficient resources in your log. Error code 998 is `ERROR_NOACCESS`. Unfortunately there doesn't seem to be much info on the internet about that error code, apparently it can refer to an access problem with the kernel driver or it can also be a more general error code. It seems to have occurred during a large write (~3mb) that requested an increase in buffer size immediately prior to the failed write:

`Dokan Information: Requesting write buffer of size 3146074.`

My guess is this has something to do with the problem. I'll see if I can reproduce the problem on my end.

Also why on earth does the user-mode driver need to request a larger write buffer? The kernel already knows how big the write buffer needs to be _before_ it forwards the request to the user-mode driver. This seems totally unnecessary to me.

Corillian:
> Also, did you rewrite the mount/unmount process ? this seems to be very fast !

Yes =)

Liryna:
>Also why on earth does the user-mode driver need to request a larger write buffer? The kernel already knows how big the write buffer needs to be before it forwards the request to the user-mode driver. This seems totally unnecessary to me.

Yes I know...I tought the same when I seen it in dokan the first time 😢 
https://github.com/dokan-dev/dokany/blob/master/dokan/write.c#L62-L63

Liryna:
@Corillian I tested Office 2016 and it "worked". I go an error at the begining that was not abel to reproduce. It was probably a write issue again.

I continue on other softwares👍 and will look on the security issue next week.

Rondom:
Thanks for the merge :-)

Corillian:
@Liryna I have been unable to reproduce the FSX race condition despite my best efforts thus far. I noticed when I merged with the latest code that there was some additional oplock code in the kernel - do you think it's possible this fixed it? Are you still able to reproduce it?

I'm working on fixing the large buffer write issue - it's taking longer than I had expected because of kernel changes.

Liryna:
Hi @Corillian,

I don't really think the oplock changes did something here but the fsx test now pass....I will try to look and see if I can still reproduce it.
I will get back to you with the result.

Liryna:
Hi @Corillian ,

I am able to reproduce the fsx appveyor error 😢 . It happen half of the runs.

Fsx output:
```
C:\Users\liryna\Desktop>fsx.exe -N 5000 M:\test
Using file M:\test
command line: fsx.exe -N 5000 M:\test
Seed set to 1476902197
truncating to largest ever: 0x29340
LOG DUMP (4 total operations):
1: SKIPPED (no operation)
2: SKIPPED (no operation)
3: TRUNCATE UP     from 0x0 (0) to 0x29340 (168768)
4: WRITE           0x388bc (231612) thru 0x3ca4b (248395)       (0x4190 (16784) bytes) HOLE
dowrite: write: Invalid argument
Correct content saved for comparison
(maybe hexdump "M:\test" vs "M:\test.fsxgood")
Seed was set to 1476902197
```

[fsxmirror.txt](https://github.com/dokan-dev/dokany/files/539695/fsxmirror.txt)

I disabled asynio here for this one:
```
#define USE_ASYNC_IO 0
```

```
C:\Users\liryna\Desktop>fsx.exe -N 5000 M:\test
Using file M:\test
command line: fsx.exe -N 5000 M:\test
Seed set to 1476901685
truncating to largest ever: 0x8078
truncating to largest ever: 0x91dc
LOG DUMP (12 total operations):
1: SKIPPED (no operation)
2: SKIPPED (no operation)
3: SKIPPED (no operation)
4: TRUNCATE UP     from 0x0 (0) to 0x8078 (32888)
5: TRUNCATE UP     from 0x8078 (32888) to 0x91dc (37340)
6: MAPWRITE        0x3080 (12416) thru 0x91db (37339)   (0x615c (24924) bytes)
7: WRITE           0x4b58 (19288) thru 0xf633 (63027)   (0xaadc (43740) bytes) EXTEND
8: READ            0x4598 (17816) thru 0x694f (26959)   (0x23b8 (9144) bytes)
9: MAPREAD         0x818 (2072) thru 0xe96f (59759)     (0xe158 (57688) bytes)
10: WRITE           0x38ecc (233164) thru 0x3ca1b (248347)      (0x3b50 (15184) bytes) HOLE
11: READ            0x26654 (157268) thru 0x30073 (196723)      (0x9a20 (39456) bytes)
12: WRITE           0x34568 (214376) thru 0x39027 (233511)      (0x4ac0 (19136) bytes)
dowrite: write: Invalid argument
Correct content saved for comparison
(maybe hexdump "M:\test" vs "M:\test.fsxgood")
Seed was set to 1476901685

```

[fsxmirror.txt](https://github.com/dokan-dev/dokany/files/539763/fsxmirror.txt)

How can I help you on this ?

Liryna:
@Corillian I just tested your last changes and was able to run fsx with more than 100k operations without issues 👍 Very good work !

Seems like you rewrote the write operation kernel side also. Is that a breaking change ?

I know that I said I will look on security descriptor issues and it is still planned. I just would like to finish this pending task for the 1.0.1 before.

Corillian:
Ok I have fixed the issues with large writes and I've also greatly simplified the code. Previously large writes worked like this:

1. Kernel check if the write is "large".
2. If no then normal write processing ensues. If yes then special handling is performed and a write IRP is created for the user driver.
3. User driver checks if a write is "large".
4. If no then normal write processing ensues. If yes then allocate extra memory and send a special write even to the kernel.
5. The kernel performed voodoo black magic to reconcile the original and new ioctl messages and returned the full write buffer.
6. Profit

The new large write logic:

1. The kernel checks if the incoming ioctl buffer is large enough to handle the outgoing buffer.
2. If yes normal logic ensues. If no then the input buffer is check to see if it is at least large enough to hold the event information. If yes then the event information is returned with an `ERROR_MORE_DATA`. If no then nothing is copied and the driver returns `ERROR_INSUFFICIENT_BUFFER`.
3. If the user driver receives `ERROR_MORE_DATA` it reads the event information, allocates a larger buffer, and re-submits the ioctl. Currently it only does this for the write IRP however it can be easily extended to encompass any IRP with the potential for exceeding the default buffer size.
4. Profit

It should be noted that currently all ioctl's are serialized through a FIFO queue in the kernel - there is no asynchronous processing. Ironically prior to my async IO changes it was possible to have multiple ioctl's in flight in the user driver due to multiple dedicated threads which meant ioctl messages could begin processing out of order. In practice this probably didn't happen often since the kernel still dequeued them in-order and each thread was blocking for the processing of its respective ioctl message *and its response ioctl*.

With my new implementation there's only ever a single dequeu ioctl in flight at any given moment however all processing of the ioctl's, including the responses, is asynchronous. If a large write is received the larger buffer is allocated and the ioctl is re-submitted. Since there is no other thread with ioctl's in flight this guarantees that the allocated buffer will in fact receive the large write message it was allocated to handle. **If this ever changes** the current code should still work though it could result in multiple threads each allocating the large buffer and racing to be the one that actually successfully receives the large write message. The losing thread would still get a valid message to process it just would have unnecessarily allocated extra memory. This could be mitigated by ensuring the ioctl message contains an ID that could be passed back to the kernel during the re-submission. Special care would need to be taken to handle the possibility of a race condition to ensure the driver returns a proper error code to the user driver if a thread requests to process an ioctl message with an ID that has already been dequeued. I believe this is actually what the old code was doing except it was using a different ioctl message type and looking up the original IRP.

@Liryna wonderful! I had a sneaking suspicion this commit might fix our issues with FSX though since I never had any problems with FSX I wasn't 100% sure. Yes unfortunately my changes to write handling are breaking changes. It's possible that there was in fact a race condition. As explained above the old code most likely would rarely hit it (though it was still possible) due to all of the blocking. I think what was happening with my new code is that it would receive a large write message, send the request back to the kernel for a larger buffer, and while waiting continue processing other ioctl messages. This would allow for multiple IO operations to begin processing before the large write message was returned. With my new changes the FIFO order out of the kernel is maintained even though the actual processing is asynchronous.

The part I still don't have enough information about is the original APC's into the kernel. Does the OS synchronize all APC's for a file or a file handle before issuing APC's or are all APC's asynchronous based on available cores, scheduling, and first come first serve? Dokan's kernel driver is currently serializing them into the queue but is that just a coincidence of how the original author decided to implement it or was it done deliberately to ensure order of processing? I'm not fully aware of what synchronization guarantees need to be made by the driver.

Liryna:
@Corillian Your write changes seem to be much cleaner than previous version and give other possibilities like you said 👍 I agree with it.

I am just wondering by your explanation if the big overlapped write are correctly handled ? I mean if there is ioctl request to create a big buffer, does still will work properly ? Or this is exactly the case you are talking ? (`Since there is no other thread with ioctl's in flight this guarantees that the allocated buffer will in fact receive the large write message it was allocated to handle.`)

About APC, there is this page that explains a "little" https://msdn.microsoft.com/fr-fr/library/windows/desktop/ms681951(v=vs.85).aspx
From what I saw, important IRP (Create, cleanup & close) are synced. It will wait that create is finished before calling other irps and close will always be at the end when other irps are completed. But for other irps like read/write, they can be made in the same time on the same handle (overlapped read/write are the best examples). You can even have them after the cleanup as you seen.

I hope I understood correctly your question ❤️

Corillian:
> I am just wondering by your explanation if the big overlapped write are correctly handled ? I mean if there is ioctl request to create a big buffer, does still will work properly ? Or this is exactly the case you are talking ? (Since there is no other thread with ioctl's in flight this guarantees that the allocated buffer will in fact receive the large write message it was allocated to handle.)

This is what I was talking about. Since all of the kernel APC's get serialized into a queue if the client (dokan.dll) only ever has a single pending ioctl then the dequeue operations are guaranteed to be serial. This means when the new large buffer is allocated and re-sends the ioctl it will attempt to read the same packet. I have tested this with large writes and my tests were successful.

Unfortunately with this design the rate that APC's are enqueued and then dequeued by the ioctl's will be the primary bottleneck. There doesn't seem to be any good way around this outside of directly forwarding kernel APC's to the user mode driver - I don't if that's possible? If it is I suspect it would require sweeping changes to both the kernel and the user mode driver.

Liryna:
I don't think there is another way to communicate with the kernel 😢 
Another possibility would be to create multiple fake device for each file but I don't think we will win speed here by doing this.

At least here with your contribution @Corillian there is no more thread limitation 👍 and that something huge ! 

I was wondering if your changes we can make STATUS_PENDING return automatically easily if the associate user thread is still busy ? In another word, remove `DokanResetTimeout`.

I will look at the security issue tomorrow 👍

Liryna:
Hi @Corillian ,

I have updated winfstest https://github.com/Liryna/winfstest. It now shows the 
```
C:\Users\liryna\Desktop\winfstest\TestSuite\t\base\10.t
not ok 2 - expect "GetFileSecurity 4cd45fc7 DACL_SECURITY_INFORMATION" <lambda> - result [{'Sddl': 'D:AI(A;ID;FA;;;WD)(A;ID;FA;;;BU)(A;ID;FA;;;SY)(A;ID;FA;;;BA)'}]
not ok 10 - expect "DeleteFile 4cd45fc7" 0 - got ERROR_ACCESS_DENIED
not ok 11 - expect "CreateDirectory 4cd45fc7 D:P(A;;GA;;;WD)" 0 - got ERROR_ALREADY_EXISTS
not ok 12 - expect "GetFileSecurity 4cd45fc7 DACL_SECURITY_INFORMATION" 0 - got ERROR_ACCESS_DENIED
not ok 13 - expect "SetFileSecurity 4cd45fc7 DACL_SECURITY_INFORMATION D:P(A;;GA;;;WD)(A;;GR;;;SY)" 0 - got ERROR_ACCESS_DENIED
not ok 14 - expect "GetFileSecurity 4cd45fc7 DACL_SECURITY_INFORMATION" 0 - got ERROR_ACCESS_DENIED
not ok 15 - expect "RemoveDirectory 4cd45fc7" 0 - got ERROR_DIRECTORY
not ok 7/15

total ................................. ok 8/15 - not ok 7/15
```

For `not ok 2`, this is because current CreateFile previously made does not use the security information parameters like before:
```
securityAttrib.lpSecurityDescriptor = EventInfo->SecurityContext.AccessState.SecurityDescriptor;
```
Instead it calls `MirrorCreateNewSecurity` -> `MirrorGetParentSecurity` that fail with access denied. Even if `MirrorGetParentSecurity` could work, we do not set the security information requested during the CreateFile.

When this will be fixed, you will have `not ok 10 - expect "DeleteFile cd13dad6" 0 - got ERROR_ACCESS_DENIED`. cd13dad6 is a file with read access refused for everyone BUT since we get access denied, we should look at the parent folder to see if we have rights to remove.

With this done, all test pass 👍 
Your Get & Set Security seems to work perfectly 👍

Corillian:
@Liryna It looks like it is in fact possible to send APC's from the kernel driver to the user driver: http://www.drdobbs.com/inside-nts-asynchronous-procedure-call/184416590

As I suspected it looks like it would require the majority of the kernel and user mode drivers to be completely rewritten. In the user mode driver it also looks like we might have to maintain our own thread pool as threads only receive APC's once they go into an alertable wait state. Needless to say such a thing is outside the scope of this PR and my available time.

As for the security stuff I'll get to that as soon as I can, thanks for following up with it!

Liryna:
Thats good to know 😃 ! 
We will probably not make this changes today I agree 👍  

Take your time 💋 but there is probably already people looking at it and specially if you find a way to "remove" `DokanResetTimeout`

Corillian:
My changes could allow for `STATUS_PENDING` to be returned to the kernel driver though I'm not sure how that would allow for the removal of `DokanResetTimeout`. Just because it returns `STATUS_PENDING` doesn't mean it can't enter a state where it never returns a result so I think you would still need to keep the timeouts in the kernel. I did however notice that I didn't update `DokanResetTimeout` for async IO so I certainly need to do that :p.

Liryna:
I had in mind something like the dokan library would "call" `DokanResetTimeout` by himself for each request / thread that is still working and going to hit the timeout. 

The timeout should only be here to dismount automatically a device that has no userland FS alive and not a limitation in the process of the request in my point of view. So if you think this could be easily added to your PR it would be nice ❤️ !

Corillian:
It can be easily added - my concern is whether or not it would be a good idea! The user driver would have to monitor the job and send keepalive packets in the background. Unfortunately if the packet deadlocked or entered some sort of unrecoverable state the keepalive background worker would have no way of knowing and would continue to send the kernel keepalive messages. I think it really should be up to the implementor of the user mode driver to determine how `DokanResetTimeout` gets called. From a usability perspective this isn't ideal however the whole point of the timeout is to prevent the kernel from infinitely consuming system resources. Also the current timeout is set to 15 seconds which would require an extraordinarily slow driver to exceed for any one operation.

Liryna:
Well 👍 OK you made your point :) 15 seconds is much enough.

Rondom:
Maybe the detection of a deadlock is over-engineering? I mean even with the new model I could have a bug in my code that continuously resets the timeout while doing nothing.

I am not familiar with how the Windows kernel handles things, but for the case that the process dies, wouldn't we able to detect that because Windows would clean up the file descriptor to the Dokan device? The automatic unmount by timeout is a useful feature for some, but my impression is that it sometimes causes more pain than it helps because people complain about "sudden unmounts" when they stress their filesystems.

Some people might even be interested in having all IO-calls block for security reasons (that's the default behaviour for FUSE on Linux) in order to not let any request through in case someone tries to circumvent some restriction by DOSing the Dokan-FS.

What are your thoughts on this? Do you think it's worth opening another issue?

martijnotto:
Are there any plans to also add fuse_lowlevel.h to the FUSE wrapper? Is this feasible now with asynchronous IO implemented?

Rondom:
The state of this PR makes me a bit sad. It is not merged because there is a yet unfixed issue with security descriptors. Unfortunately, I am not qualified enough to know how to fix it. I see master diverging more and more making it more difficult to merge the changes. Maybe developing on this branch for the next major version can accelerate things and give a wider exposure, because people will test pre-release versions? Anyways, this does not solve the actual problem of missing manpower...

@martijnotto If you look closely at the FUSE low-level API you can see that it is quite different from Windows semantics, i.e. it is a bit of work to implement translating between those semantics in a somewhat efficient manner. For instance, Windows works with full paths, while low-level FUSE works worth with the parent-dir (inode) and filename.

Without having tried it, I think this PR might indeed make it possible or at least easier to implement such a wrapper. Finding out how well it works and what other obstacles there are requires one to actually start investigating/trying it.

Liryna:
I agree @Rondom This PR is really great and is missing manpower 😢 
It could really enhance dokan on the speed 👍 I have not forgotten it but I have no more free time available to give.
We need someone here. If anyone would it and have the resource to do it, fill free, I would be happy to help and guide.

martijnotto:
@Rondom wouldn't it be enough to simply keep a hashmap of fullpath => inode. I believe that is what the FUSE high level API does too. As to other differences, I have no idea what those are exactly since I am not at all familiar with Windows. I am simply looking at the easiest way to get my app to work on Windows too when it is finished.

Rondom:
It would, probably. I expect quite a few edge cases to take care of in general. Not saying that it's impossible, but it might need some time to get something like this completely right.

First, someone needs to fix the last bugs here and get this merged, though...

Liryna:
Hi @Rondom ,

I have create a new branch and merged @Corillian changes https://github.com/dokan-dev/dokany/tree/Corillian-asyncio
I have not fully tested but the merged seems to have worked 👍 I would probably need your help for fuse.
I had some issue to merge https://github.com/dokan-dev/dokany/commit/f113f7476efdfd674791baa72bc30a503e879e2a and I am not sure I did it correctly. Can you just take a look a it ? ❤️

Rondom:
> I have created this branch to merge all changes of the master into it to keep it update.Love @Corillian changes ❤️ speed has really been improved !

This is great news! Thanks a lot for reviving this branch (and of course thanks again to @Corillian 

> I had some issue to merge f113f74 and I am not sure I did it correctly. Can you just take a look at it ? 

Sure, will do as soon as time allows.

I will also try to work further on our continuous integration scripts to get more tests running and possibly do some baselining against master to see whether there are any regressions. I cannot give a timeframe on that, though.

Liryna:
**TODO** here:
- [ ] See if https://github.com/dokan-dev/dokany/issues/393 happen
- [ ] Check correct merge of https://github.com/dokan-dev/dokany/commit/f113f7476efdfd674791baa72bc30a503e879e2a
- [ ] @Corillian said `I did however notice that I didn't update DokanResetTimeout for async IO so I certainly need to do that :p.` This should be changed
- [ ] Fix security [issues](https://github.com/dokan-dev/dokany/pull/307#issuecomment-257590762)
- [ ] Heavy test ( @Rondom ifstest ?) .
- [ ] Update documentation

magol:
@Liryna
I also want to add this to the TODO list:
  - [ ] Update [Dokan-DotNet](https://github.com/dokan-dev/dokan-dotnet ) to work with the async API.

That should of course be at the bottom of the list when the other are fixed. My question is what should be done before there is any idea to start looking at the C# version?

Liryna:
@magol We will need to convert all struct that now holds all operations params.
https://github.com/dokan-dev/dokany/pull/307/files#diff-641844c42f5ba7de1b7fd5347b56b3eeR245
After we just need to change the all operations params `foo` as `barStruct->foo`
Even for C# we will probably even not break compatibility if we want. Keep the new changes in the operationProxy.

magol:
Great

If we have to guess vild, what is the time frames? When do we think it's ready enough to start beta testing (and when we can start port Dokan-DotNet)?

Rondom:
@Liryna That sounds like a realistic (and manageable!) list. I always 
try to aim for "too perfect", thus making things take very long, which 
is not very good. :-)
After letting this PR linger for a while I can understand the desire to 
get this into the hands of people quick (while still doing enough 
testing, of course).

What do you think about implementing a small wrapper that enables using 
the old API? At least many of the open-source projects using Dokan are 
very short-staffed and it might take some time until they adopt a new 
API. That way we can mark the new API experimental and still do some 
other API-breaking changes along the way and thus avoid to break the API 
too often.
One or two releases later we mark the new API as stable and deprecate or 
eventually remove support for the old API.

Question is: Do you think that's worth the effort (both implementation 
and maintenance of the old API for at least one or two releases)? I 
would volunteer to give it a try. I would love to hear your opinions.
It might as well be not worth it and I am underestimating the effort 
needed for on-going testing and support.

Actually after having written down all this I am more skeptical than 
ever before and it's probably better to collect a list of probably 
API-breaking changes and (if time allows) strive to implement the 
low-hanging fruit before the final stable release. I am still leaving 
the rest of the text above here to illustrate my thought process...

Liryna:
@Rondom Having the old API work for a couple of release is probably possible yes.
Will have to see how to recreate the DokanMain entry with `DokanInit()` and make it not async.

We will do a Wiki page to explain how to convert 1.x.x to 2.x.x when the moment comes 👍

Rondom:
@Liryna I have looked at the merge of the FUSE-change and it looks fine, but honestly I am not really confident by just looking at the code. Ideally someone with NodeJS-installed can test (or even better we would have a test-case for that......).

I thought I would find the time to do some fix for some other issue over the weekend, but I did not. Instead I am reporting issue #511 for the time being. If someone starts working on it, please write in the issue. I will do the same in case I find the time to avoid duplicating work.

Liryna:
Here is first BETA: https://github.com/dokan-dev/dokany/releases/tag/v2.0.0-BETA1 👍

Daniel-Abrecht:
I've compiled fuse-nfs with the v2.0.0-BETA1 version, just to see if it works or not. It turned out the -s option to disable mutithreading in fuse does no longer work in dokanys fuse wrapper. Single threading is often prefered in programs which use fuse, because it prevents any race conditions, keeps things simpler, doesn't require all used libraries to be thread safe and isn't really slower for a lot of use cases. Fuse-nfs and probably a lot of other programs which use fuse relay on its single thread mode and can't be used without it.

Corillian:
Are you suggesting that every single file system event be serialized to a single thread?

Daniel-Abrecht:
Yes, but only if the program which uses fuse requests the single thread mode. The current fuse wrapper tries to do this here: 
https://github.com/dokan-dev/dokany/blob/af49c7200fefe9d46d06a00f9413b466142af746/dokan_fuse/src/dokanfuse.cpp#L604

Alternatively, it would probably be sufficent to just use a mutex lock to prevent two fuse callbacks to be called at the same time. Or if the single thread mode won't be supported anymore, there should at least be a warning if a program attemps to use it.

Corillian:
Excluding debugging I can't think of any reason why you would want to do that regardless of whether or not you're using FUSE. It would completely destroy the performance of your filesystem. That being said, and as you suggested, there's nothing preventing you from forcing your FUSE driver from being single threaded by taking a lock on a mutex.

Daniel-Abrecht:
I already gave a lot of reasons way a lot of fuse file systems prefer single thread mode over multithreaded mode. There are also a lot of cases where not using multithreading can actually increase the overall system performance. Let me give you an example: Let's assume we have a network file system which uses a single socket to transfer datas. In that case, the network would be the bottle-neck, and having multiple threads which try to send data and/or synchronize some actions just wastes cpu time other processes could have used.

However, I'm not interested in the single thread mode because of it's possible bennefits and drawbacks. 
The interesting thing about fuse is that it's an api that allows a program to provide a file system and that program will work on any system which has an implementation of the fuse api. The single user mode of fuse is a basic feature every implementation of fuse provides and many programs rely on. Reimplementing this feature in an existing fuse file system which previously relayed on the fuse implementation to provide it is somthing I would never do, it wouldn't make any sense.

Just to make this clear, this isn't about just adding a feature I think may be neat, this is about maintaining compatiblity with existing programs and the fuse api itself.

Corillian:
The examples you have provided would be _much_ less performant on a single thread instead of using multiple threads - _particularly for a networked filesystem_. From a performance perspective there are no benefits to using a single thread which is why Windows uses APC's to talk to the kernel driver in the first place. I'm not familiar with the gritty details of building a filesystem for Linux however, from what I remember of working on the FUSE API for this PR, it appears to be poorly designed because each call to the FUSE API must block a thread until the requested operation completes (unless you implement coroutines). If what you are saying was true Windows wouldn't support APC's, overlapped IO, or IO completion ports and Linux wouldn't support polling.

`ThreadCount` is no longer used because Dokan's user mode driver now uses the system thread pool which creates and destroys threads on the fly based on the number of CPU's in the system along with other heuristics. I probably should have just removed it. If FUSE supports forcing everything to run through a single thread and a flag was added to enforce such behavior in Dokan for compatibility purposes I certainly wouldn't complain - though you'd be crazy to use it in production.

Liryna:
Well point @Daniel-Abrecht , I forgot about the single thread option of FUSE.
@Corillian crazy as it is, I already see a couple of FUSE fs using the single thread option because of shared resources... 😢 Will need to add a way to limit the thread pool to 1 for compatibility.
People already having their FUSE FS are not always willing to change it for windows even if it is a simple mutex to add.

martijnotto:
I don't currently use dokany, but am keeping a close eye on it, particularly on asynchronous IO and on the hopeful implementation of the FUSE lowlevel API as this would make the eventual porting of my application a whole lot easier.

My application currently uses a single thread and is centered around the libuv event loop. Since it is a networked filesystem, and the socket is the bottleneck, additional threads don't add any extra performance. Rather, they'd take more resources and slow things down due to the necessity of working with mutexes.

The asynchronous IO feature is really neat, but forcing people to use multiple threads is - IMHO - silly.

lostmsu:
Just came here for the first time. Is this pull request still up to date? Or was it reworked in some way? E.g. where can I find the diff.

Liryna:
@lostmsu The PR has been updated here:
https://github.com/dokan-dev/dokany/tree/Corillian-asyncio

Only 11 commit missing compared to master. Will do the update during the weekend 👍

timofonic:
@Corillian @Liryna @Rondom @Daniel-Abrecht @marinkobabic @lostmsu 

Any news about this?

@Liryna If this PR has been updated in another fork. Why not upate this one or close and create a new one?

I'm not a developer, but I get confused by so many somewhat abandoned PRs :P

Liryna:
@timofonic This PR is keep as a reference of Corillian since he is lacking of time to continue.

However, dokany has his own branch updated with last changes.
Also i keep the pr open for the history and discussion.

But you are perhaps right, i should rename the branch for v2 and create separate issue for the v2 and roadmap! :)

Liryna:
This is now partially merged in 2.0.0. the only remaining part I see that could be interesting to merge is the possibility for userland to return status pending and call the dokan API when the operation can be completed.

Issue #318, opened 2016-08-19
mirror: when copy more than 65000 files => BufferOverflow or OutOfMemory or Deadlock


ed-miller:
Feature request can skip this form. Bug report **must** complete it. `Check List` must be 100% match or it will be automatically closed without further discussion. Please remove this line.

# Environment
 * Windows version: Win7 x64 Pro
 * Processor architecture: x64 Intel i7 4790, 4GB RAM, 512G SSD
 * Dokany version: 1.0.0.4000
 * Library type (Dokany/FUSE): Dokany / Dokany.sys / Dokany.dll / Mirror.exe
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description
When copy in Windows Explorer C:\ to Mirror O:\ => S:\ (independent of path) a huge amount of small files - in my case about 67000 files - then the dokan.sys or dokan.dll hangs up (Deadlock?). Often Windows Explorer Progress Dialog show message ERROR_INVALID_PARAMETER or E_OUTOFMEMORY before it deadlocks.
I have repeated the test over 30 times. Different Mirror targets sync paths. Different Mirror MountPoints. Different files types or sizes. Every time the same.

# Logs
Loging not possible. Hundred thousends of debug lines. Debug view hangs up

marinkobabic:
Please enable the driver debugging at the position when you think you will get the deadlock  and not from beginning on.

This would reduce the log size.

Liryna:
Fixed with https://github.com/dokan-dev/dokany/pull/325

PR #325, opened 2016-08-31
Rewrite FCB locking

taruti:
This is unfortunately somewhat large.

Tested with copying deep filetrees with 80000+ files.

Related to:
https://github.com/dokan-dev/dokany/issues/318
https://github.com/keybase/client/issues/4018

### DIFF
--- a/sys/cleanup.c
+++ b/sys/cleanup.c
@@ -33,7 +33,7 @@ Routine Description:
 Arguments:
 
         DeviceObject - Context for the activity.
-        Irp 		 - The device control argument block.
+        Irp          - The device control argument block.
 
 Return Value:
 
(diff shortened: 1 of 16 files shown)


Liryna:
Hi @taruti ,

So this issue were only related to fcb lock that was not properly handled ?
Can you explain a little what was happening :heart: 

I will try to test it quickly.

taruti:
There were rare BSODs with backtraces hinting to lacking locking of FCBs typically with a huge amount of file operations.

The existing locking code was not comprehensive and lacking in some cases. After trying to add incremental changes to fix it I ended up with just rewriting and adding locking for FCBs.

Liryna:
:+1:  Have been able to copy 260k files without issues !
Lock seems to be well placed and are always release as I can see.

Well done @taruti and thanks again, very appreciate as always ! :trophy:

PR #326, opened 2016-09-01
sys: FCB unlock on some error paths

taruti:
Found some error paths that were still lacking proper locks when reviewing the code for the third time.

### DIFF
--- a/sys/cleanup.c
+++ b/sys/cleanup.c
@@ -92,6 +92,7 @@ Return Value:
 
     if (eventContext == NULL) {
       status = STATUS_INSUFFICIENT_RESOURCES;
+      DokanFCBUnlock(fcb);
       __leave;
     }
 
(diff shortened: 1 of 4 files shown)


taruti:
After this is in I'll do a review of CCB locking.

No known crashes on that one, but better catch it before any mysterious issues arrive. It will be much smaller than the FCB one.

Liryna:
Thank you @taruti and :+1: well found !

Is there not also one missing here ? (not really important since it is for code before Win7)
https://github.com/keybase/dokany/blob/e309f53feed6531a2348a05ff549b1653c056f35/sys/create.c#L1079

I will try to review other places also.

A review of CCB would be very appricate also !!! :+1:

taruti:
No, in that case the return on line 1079 is inside the __try and it is handled in the longish __finally clause starting on line 1171 and the unlock on line 1203.

Liryna:
Alright @taruti I missed it :+1: 

I have reviewed again all FCB, they seems all well fixed with your last pull request and this PR.
Just this one: https://github.com/dokan-dev/dokany/blob/424692c0d36af2338a7af5f5d5000ff344e05e63/sys/cleanup.c#L117
There is a __leave line 95, should we not unlock it before leave or in finally ?

Good work @taruti :trophy:

taruti:
Yes, updated the PR and it handles that.

Thanks for taking time to review this!

Liryna:
Sorry ! It was in front of my face :smile: 

Thanks again for your contribution !

PR #327, opened 2016-09-01
sys: Use FCB locks for CCBs

taruti:
This might be somewhat controversial - it simplifies some things - and requires less locks - but might be pessimistic in cases with a lot of lock contention.

There is no hurry to merge this and doing this in another way might be sensible :)

Should we use separate CCB locks?

Merging the locks means:
+ Somewhat simpler code
+ Somewhat smaller CCBs and less locks around
+ No need to be careful to take FCB and CCB locks in the right sequence
+ In places where both  FCB and CCB lock would be used just the FCB lock suffices
+ Which can be faster if there is not very much contention
- We take an FCB lock in a few places where a CCB specific one would suffice
- Which can be slower if there is lots of contention

I have left UserContext alone with this patch.

Looking forward to discuss this :)

### DIFF
--- a/sys/create.c
+++ b/sys/create.c
@@ -226,7 +226,7 @@ PDokanCCB DokanAllocateCCB(__in PDokanDCB Dcb, __in PDokanFCB Fcb) {
 
   ccb->Fcb = Fcb;
   DDbgPrint("   Allocated CCB \n");
-  ExInitializeResourceLite(&ccb->Resource);
+  //ExInitializeResourceLite(&ccb->Resource);
 
   InitializeListHead(&ccb->NextCCB);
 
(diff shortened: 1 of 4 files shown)


Liryna:
I even question myself if ccb lock are needed. There is already a couple a place where ccb is used and not locked. (like https://github.com/dokan-dev/dokany/blob/424692c0d36af2338a7af5f5d5000ff344e05e63/sys/fileinfo.c#L237)

As far as I know, ccb is only used in the current request/context and there should not be another thread using the same ccb. @marinkobabic can you confirme what I say or correct me if am wrong :heart: ?

taruti:
Don't they use a global lock (_Global_critical_region_) in that example for most things?

Ccbs are created in `DokanDispatchCreate` and free'ed in `DokanDispatchClose`. Even if an application does two syscalls on the same handle from different threads concurrently things should work in the driver.

marinkobabic:
The lifetime of fcb is much bigger than the one of the ccb. Also the fcb is more global than an ccb. Blocking fcb because of changes on ccb would be inefficient. Threads would unnecessary be blocked. Its like blocking the whole house because of some changes in a room instead of blocking just the room itself.

taruti:
So do things with per-CCB locks? I can make a patch that does exactly that if wanted.

marinkobabic:
@taruti 
Correct

PR #328, opened 2016-09-01
check for GDR/LDR version of KB3033929 in bundle

zanderz:
@Rondom suggests this check may be helpful in the bundle, not just the individual packages.

### DIFF
--- a/dokan_wix/Bootstrapper/Bundle.wxs
+++ b/dokan_wix/Bootstrapper/Bundle.wxs
@@ -15,7 +15,7 @@
             UpdateUrl="$(var.UpdateURL)"
             DisableModify="yes"
             >
-
+      <util:FileSearchRef Id='WINTRUST_FileSearch' />
       <BootstrapperApplicationRef Id="WixStandardBootstrapperApplication.RtfLicense">
         <bal:WixStandardBootstrapperApplication
           LicenseFile="Resources\licdata.rtf"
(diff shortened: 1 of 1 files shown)


Liryna:
Hi @zanderz ,

Is it not the same as https://github.com/dokan-dev/dokany/commit/9782a6491775ef3660a71f40f381d3f48e511f14 ?

Rondom:
No. Without this, the bundle will happily install all kinds of stuff like .NET only to then fail while installing the MSI at the end. If I remember my tests correctly at least .NET installation is not even rolled back. With this patch, we fail early.

Liryna:
Alright @Rondom ! I agree that it would give a better experience to add this check before.

Does the check should not be 
((WINTRUSTVERSION >= v6.1.7601.18741 AND WINTRUSTVERSION `<` v6.1.7601.22000) OR WINTRUSTVERSION >= v6.1.7601.229`48`)) ?

Rondom:
Depends on whether the bundle-engine follows arcane MSI implementation details that make one pull their hair out :-D

js69:
There is a minor text bug: " ... the installation of Keybase."

Rondom:
I would guess you do not need to decrease the version number by one, because here the version number is stored in a variable and I would expect at least basic operators like greater, lesser and equal would work in a sane way. Then again you don't know whether the right number gets stored I the variable... Googling made me none the wiser. In the worst case I will have to test with both the GDR and LDR patch...

Rondom:
http://wixtoolset.org/documentation/manual/v3/xsd/util/filesearch.html
^ this suggests that the bundle behaviour is sane i.e. no decrementing needed

zanderz:
Whoops, sorry about the product name, I fixed that.
Our reason for moving this logic out of the package and into the bundle is for our own bundle to show coherent errors to the user. At first, our bundle included the Dokan bundle in its chain, and when run silently, all its errors were masked, and worst of all, the visual studio redistributable package could cause an unprompted reboot. Now we have copied all the same prerequisites from the Dokan bundle to ours, and just chained the .msi packages.

Speaking of which: it would be very helpful if the .msi packages were included in the release binaries, so we hopefully someday can stop building + signing our own.

Liryna:
@Rondom I agree with you, just that Wintrust.dll has version 6.1.7601.18741 and 6.1.7601.22948
https://support.microsoft.com/en-us/kb/3033929
So the check should be `WINTRUSTVERSION >= v6.1.7601.22948` and since LDR start as 22xxx the check should be `WINTRUSTVERSION < v6.1.7601.22000`.

@zanderz :+1: for the msi packages ! I added it to the build script (it remind me to upload it :smiley: ) https://github.com/dokan-dev/dokany/commit/8885fefe75e1aaa299cf94ef19a70428c396cafb

Liryna:
@zanderz I am ready to merge your PR, I am just waiting for the changes on the version value as describe before (correct me if I am not wrong in my way I think this work :smile: ) and also rebase into a single commit.

zanderz:
Changed to `WINTRUSTVERSION < v6.1.7601.22000` and squashed down.

zanderz:
I changed it, but note that it doesn't match the packages, because MinVersion is inclusive.

Liryna:
>I changed it, but note that it doesn't match the packages, because MinVersion is inclusive.

Here it is directly the correct version of the wintrust that we compare, there is no MinVersion.

I tested it on Win7 with and without GDR/LDR KB and got the correct message just before dependency get installed :+1: 

Good job @zanderz and very appreciate !

PR #336, opened 2016-09-14
Appveyor: Upgrade all Cygwin packages before build

### LINKED ISSUE #335: Cygwin-cmake is broken on Appveyor-CI
Currently we always install the latest version of cmake for Cygwin when doing CI builds on Appveyor.
For some reason cmake is broken on Appveyor.

I logged in via RDP to the build worker and found out that the cmake command will just silently exit, no matter what parameters you provide.
Reinstalling cmake using the Cygwin installer seems to fix the problem as far as I can see.

Does anyone have any ideas what is wrong or want to troubleshoot themselves?

Rondom:
Upgrading all packages ensures that we are running the latest Cygwin.
Sometimes dependencies of packages are incorrect and dependant packages
are thus not upgraded automatically when they should be.

This makes the Appveyor build working again after the cmake-package in
Cygwin was upgraded a few days ago breaking our build. Closes #335.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -36,7 +36,7 @@
           
           function updateCygwin($cygwinexe, $installFolder, $cacheFolder) {
             Write-Host "Update Cygwin: " $cygwinexe -NoNewline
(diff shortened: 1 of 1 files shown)


Rondom:
works :-)

Liryna:
:+1: perfect thank you @Rondom !!!

PR #347, opened 2016-09-22
sys: Do not hold locks for CcPurgeCacheSection

taruti:
Rearrange locking  so that the FCB lock is not held for `CcPurgeCacheSection`.

Quick fix - requires testing.


Liryna:
@taruti work well on my side 👍 Thank you for the quick fix ! very appreciate and well found !

PR #348, opened 2016-09-23
Use atomic operations for FCB and CCB flags instead of locks

taruti:
* Go through all Flags access for FCBs and CCBs
* Use helpers for access
* Make modification of flags use atomic operations
* Locks need to be held for less time because they are not needed for flags anymore


Liryna:
Hello @taruti ,

Thanks you for the contribution !
This is a good improvement of how the flags are checked (cleaner, same macro used everywhere) and the safety of it !

Very appreciate ! keep the good work ❤️ !!!

Issue #338, opened 2016-09-14
dokan_fuse: Ship pkg-config file


Rondom:
We might want to ship a pkg-config file to simplify building against dokanfuse. At least during the cmake-install this is easy to accomplish.

http://www.kaizou.org/2014/11/typical-cmake-project/

I am not sure whether we can ship it with the installer, though. Reason: One can use pkg-config for compiling under Cygwin, MSYS or Windows. All three have different means of specifying the same path (`C:\foo`,`/c/foo`,`/cygdrive/c/foo`) and the LDFLAGS may also be different

Rondom:
The instructions in the link above can be simplified by not using so many custom variables (`PKG_CONFIG_`). For example one can reuse `CMAKE_INSTALL_INCLUDEDIR` instead of defining a separate custom variable `PKG_CONFIG_INCLUDEDIR`.

In my opinion, we should probably not use the `fuse` name for our pkg-config-file. There are several different user space file system solution and our solution is not the original libfuse. We should respect the original libfuse (unlikely they ever accept patches bringing Windows-compatibility, but still) and also avoid a conflict with any alternative user-space FS solutions. I would like to hear other people's opinions on this.

Rondom:
I opened PR #350.
It needs testing and some discussion about how to handle the fact that we are using the libfuse-name.

Rondom:
I opened PR #375.
It needs testing and some discussion about how to handle the fact that we are using the libfuse-name.

PR #350, opened 2016-09-24
Msys2 pkg-config

USeebi:
Msys2 fuse library pkg-config and pacman package to be consumed by
makepkg


Liryna:
Hi @USeebi ,

Thank you for the pull request.

Is it not possible to have a pkg-config that copy all headers in the correct folder from the dokan_fuse project rather than having duplicate headers in the repository ?

Rondom:
Hi @USeebi,
thanks for helping making Dokany better.

# pkg-config
Nice to see that you added a pkg-config file! :+1:  We actually already had an enhancement request about it: #338 
pkg-config is not specific to MSYS2, but actually generic to any platform. Therefore it would be nice to have it available for any platform (MSYS2, Cygwin, Linux (for cross-compilation)) and thus independent of the library and include directories.
Do you think you can integrate the pkg-config into our cmake-build using the instruction in the ticket linked above? You can create a separate Pull-Request for this, to simplify matters.

# MSYS2-PKGBUILD:
First of all: Is there a specific reason, why you are submitting this to us and not to @Alexpux/MSYS2-packages? I am not particularly opposed to having the PKGBUILD in the Dokany repository,  but I would be interested in what your considerations where and why you prefer one over the other option.

I have never done any packaging for MSYS or Arch Linux, so take the following comments with a grain of salt. (I am mostly familiar with Debian packaging and Linux and MSYS2 might differ in several ways I don't know yet about).

As @Liryna has said, you duplicate all header files in the repository. Looking at your packaging, it seems quite manual and I think it can be improved. The way I know it from Debian or RPM-based distros, the package would get built and installed into a separate directory as if you were compiling/installing from source and then a package is created from that.

I have just searched for "cmake" in the MSYS-packages repository and the doxygen-package there might be a good example of cmake-based project that is packaged using a PKGBUILD where the source is automatically downloaded, verified using hashes, built and "installed" to be compressed into a package.
https://github.com/Alexpux/MSYS2-packages/blob/45df194fc83273dec06cdba2497c243a50883a40/doxygen/PKGBUILD

Some questions:
1. Where is the library-file built/installed to in your current implementation?
2. Why do you install the headers to `/opt/`? That seems very non-standard to me for packaged software. Why not `/usr/include/`?

PS: Thanks again for your contribution, it's appreciated.

USeebi:
Hi Rondom,

You're right, the package is manually pulled together after compiling dokan_fuse for makepkg to work with dependency on fuse.  I guess the package mostly working with Msys2 as Mingw has no syslog while Linux probably use fuse directly but not dokan-fuse.

The latest Msys2 distribution automatically set /opt/bin in the path so that I put the dokan_fuse library in /opt/bin and fuse include files in /opt/include/dokan_fuse.  The installation is automatic with pacman although the package is manual and the /opt directory might be a good place to signify the non-standard.

All these information are coded in fuse.pc and this could be automated with Autotools while there are also PKGBUILD templates with Autotools.  I'm not really expert with Autotools and I really need to brush up with Cmake.  I guess we need someone expert in both Autotools and Cmake to merge this pull request at will.

USeebi:
An other way to automate is manipulating build.ps1 script 

& dokan_fuse/cleancmake.bat
Write-Host "Build Msys2 x64..."
& C:\msys64\msys2_shell.cmd -msys2 -lc "cd '$currentPath'/dokan_fuse/ && cmake . -G 'MSYS Makefiles' && make" | Out-Host
New-Item -Force -Type Directory "pkg/opt/bin/", "pkg/opt/include/", "pkg/usr/lib/pkgconfig/"
Copy-Item "dokan_fuse/fuse.pc" "pkg/usr/lib/pkgconfig/"
Copy-Item "dokan_fuse/.PKGINFO" "pkg/"
Copy-Item "dokan_fuse/libdokanfuse1.dll" "pkg/opt/bin/"
Copy-Item "dokan_fuse/libdokanfuse1.dll.a" "pkg/opt/bin/"
Copy-Item -Recurse "dokan_fuse/include/" "pkg/opt/include/dokan_fuse/"
Remove-Item "dokan_fuse/libdokanfuse1.dll"
Remove-Item "dokan_fuse/libdokanfuse1.dll.a"
& C:\msys64\msys2_shell.cmd -msys2 -lc "cd '$currentPath'/dokan_fuse/pkg && tar -cf - .PKGINFO * | xz -c -z - > ../dokan_fuse-x86_64.pkg.tar.xz"
Write-Host "Build Msys2 x64 - Done" -ForegroundColor Green
Write-Host "Install package with pacman -U dokan_fuse-x86_64.pkg.tar.xz" -ForegroundColor Green

These are the steps that I pulled up the package and I would like to insert the dokany version into the package file name but need more efforts.

Rondom:
@USeebi May I ask, why you need an *MSYS2*-package? As far as I am informed, MSYS2 is a fork of Cygwin and its only purpose is to serve as a POSIXy environment for building MinGW-W64 (32bit and 64bit) applications. So one is not supposed to build software using the MSYS2-GCC if one is not developing MSYS2. If one needs an environment more close to Unix, the MSYS2-people suggest to use Cygwin.

I created pull-request #375 that enhances our build-system to automatically to create a pkg-config-file depending on the installation path. It needs some testing ;-)

Are you interested in working further on this to create an automated package that avoids copying files around? Like the libssh2 or doxygen-package I suggested above? Maybe, once a version with my PR merged is released and the package has good enough quality, you could even submit it to [@Alexpux/MINGW-packages](https://github.com/Alexpux/MINGW-packages/)?

Rondom:
>  I guess the package mostly working with Msys2 as Mingw has no syslog while Linux probably use fuse directly but not dokan-fuse.

Can you please explain what you mean here in more detail? I do not really understand 😕

timofonic:
@USeebi @Liryna @Rondom 

Any news about this? Any reason this is open? Can someone reply about this? Thanks in advance :)

Issue #359, opened 2016-10-01
Dokan causes BSOD on filesystem mount on Windows 10 build 14936


cerebrate:
# Environment
 * Windows version: 10 (Insider build 14936)
 * Processor architecture: x64
 * Dokany version: 1.0.0 release (latest)
 * Library type (Dokany/FUSE):  Dokany
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description

System blue-screens with SYSTEM_SERVICE_EXCEPTION as soon as any user-mode filesystem (tested with keybase, mirror, own test code) attempts to mount.

# Logs

Memory dump attached here:

https://1drv.ms/u/s!AqrcZdowvFOLqLsbQypTta7rom317w

Liryna:
Hi @cerebrate ,

Thank you for the report !

I cannot use the dump unfortunately :'( is not me who built the version (need the dbg from the build).

This is a very interesting case to see what Microsoft changed! I will follow the KeyBase discussion 👍

cerebrate:
@Liryna Is there a way I can get you a dump that you can use? (If it's just a matter of using the debug version of Dokan as per the "How to Debug" wiki page, I can do that easily enough.)

Liryna:
@cerebrate try to install https://github.com/dokan-dev/dokany/releases/download/v1.0.0/DokanSetup_redist-1.0.0.5000.exe (enable optino install dev tools). Run the mirror in dokan sample folder.
`mirror.exe /r C:\Users /l m`

Browse the M:\ 👍

CrosseyeJack:
I tried to follow the instrustions on the wiki and installed your advice to install redist which dev tools but that didn't give me mirror.exe to run so I downloaded it from the release zip. Ran dbgview, told it to log to a file, ran the command and it BSOD'ed and wrote nothing to the log file. tried to capture stdout&stderr but again no output was written before BSOD.

If I'm doing something wrong please let me know and i'll try again.

Liryna:
[...]

PR #363, opened 2016-10-06
fix bsod issue related to filesystem mount on Windows 10 build 14936

marinkobabic:
Fix for issue https://github.com/dokan-dev/dokany/issues/359. Please do a review if I'm doing the buffer copy correct. Maybe is there a much simpler way to compare the strings?

The issue is related to the dokan architecture. If we face more such problems in the future a generic approach must be implement to avoid such problems.

### DIFF
--- a/sys/create.c
+++ b/sys/create.c
@@ -528,6 +528,27 @@ Return Value:
     if (!vcb->HasEventWait) {
       DDbgPrint("  Here we only go in if some antivirus software tries to "
                 "create files before startup is finished.\n");
+      if (fileObject->FileName.Length > 0) {
+          DDbgPrint("  Verify if the system tries to access System Volume\n");
+          UNICODE_STRING systemVolume;
+          RtlInitUnicodeString(&systemVolume, L"\\System Volume Information");
+          if (fileObject->FileName.Length >= systemVolume.Length) {
+              LPCWSTR p1, p2;
+              p1 = systemVolume.Buffer;
+              p2 = fileObject->FileName.Buffer;
+              unsigned int len = 0;
+              LONG ret = 0;
+              while (!ret && len < 26) {
+                  ret = *p1++ - *p2++;
+                  ++len;
+              }
+              if (ret == 0 && (systemVolume.Length == fileObject->FileName.Length || *p2 == L'\\')) {
+                  DDbgPrint("  It's an access to System Volume, so don't return SUCCESS. We don't have one.")
+                  status = STATUS_NO_SUCH_FILE;
+                  __leave;
+              }
+          }
+      }
       status = STATUS_SUCCESS;
       __leave;
     }


Liryna:
Hi @marinkobabic ,

Well found as always and extremely appreciate ❤️ !

About unicode string compare, can we not just call RtlCompareUnicodeString with fileObject->FileName and systemVolume, check the return value like:

```
RtlInitUnicodeString(&systemVolume, L"\\System Volume Information");
if (RtlCompareUnicodeString(&systemVolume, &(fileObject->FileName), TRUE) < 0) {
   DDbgPrint("  It's an access to System Volume, so don't return SUCCESS. We don't have one.")
   status = STATUS_NO_SUCH_FILE;
   __leave;
}
```
it would avoid the buffer copy.

RtlCompareUnicodeString doc says that `< Zero : String1 is less than String2.`, so it "should" be alright 😄 do you think you can test something like that ?

marinkobabic:
What we need is the check if the path starts with \System Volume Information

The passed in path is 
\System Volume Information\Cloud Files\

Liryna:
Yes, I tought `RtlCompareUnicodeString` would return a <= 0 value if `fileObject->FileName` started with `\System Volume Information`.
I tested it and I was wrong 😢 

```
    UNICODE_STRING string1, string2, string3, string4;
    RtlInitUnicodeString(&string1, L"\\System Volume Information");
    RtlInitUnicodeString(&string2, L"\\System Volume Information\\Cloud Files\\");
    LONG result = RtlCompareUnicodeString(&string1, &string2, TRUE);
    DDbgPrint("RtlCompareUnicodeString 1 result %ld\n", result);

    RtlInitUnicodeString(&string3, L"\\System Volume Information2\\Cloud Files\\");
    result = RtlCompareUnicodeString(&string1, &string3, TRUE);
    DDbgPrint("RtlCompareUnicodeString 2 result %ld\n", result);

    RtlInitUnicodeString(&string3, L"\\test");
    result = RtlCompareUnicodeString(&string1, &string4, TRUE);
    DDbgPrint("RtlCompareUnicodeString 3 result %ld\n", result);


.....

[DokanFS] RtlCompareUnicodeString 1 result -13
[DokanFS] RtlCompareUnicodeString 2 result -14
[DokanFS] RtlCompareUnicodeString 3 result 26
```

BUT this worked:

```
    UNICODE_STRING systemVolume;
    RtlInitUnicodeString(&systemVolume, L"\\System Volume Information");

    if (systemVolume.Length >= fileObject->FileName.Length)
    {
      LPCWSTR p1, p2;
      p1 = systemVolume.Buffer;
      p2 = fileObject->FileName.Buffer;
      unsigned int len = 0;
      LONG ret = 0;
      while (!ret && len < 26) {
          ret = *p1++ - *p2++;
          ++len;
      }
      if (ret == 0 && (systemVolume.Length == fileObject->FileName.Length || *p2 == L'\\'))  {
        DDbgPrint("  It's an access to System Volume, so don't return SUCCESS. We don't have one.")
          status = STATUS_NO_SUCH_FILE;
        __leave;
      }
    }
```
it is not much but...we avoid allocating memory 😃 

Could you try it on your side if still work and update your PR ❤️ ?

marinkobabic:
@Liryna 
Have done the requested changes and it works fine 👍 Nice solution to avoid memory allocation 😄 
So let's hope how that the BSOD can't be reproduced anymore. Thanks a lot!

Liryna:
Thank you for the quick changes ❤️ !

PR #370, opened 2016-10-17
MirrorCreateFile - Reject when trying to open a file as a directory

Kerbox:
STATUS_NOT_A_DIRECTORY should be returned if a file is being opened and the FILE_DIRECTORY_FILE option is used.
The problem with allowing this, is that if the file has previously been opened as a normal file, it will set the (shared) FCB as being a directory ([here](https://github.com/dokan-dev/dokany/blob/master/sys/create.c#L1360)). As such, further reads on the file will fail!

### DIFF
--- a/samples/dokan_mirror/mirror.c
+++ b/samples/dokan_mirror/mirror.c
@@ -326,6 +326,13 @@ MirrorCreateFile(LPCWSTR FileName, PDOKAN_IO_SECURITY_CONTEXT SecurityContext,
       }
     }
     if (status == STATUS_SUCCESS) {
+      //Check first if we're trying to open a file as a directory.
+      if (fileAttr != INVALID_FILE_ATTRIBUTES &&
+          !(fileAttr & FILE_ATTRIBUTE_DIRECTORY) &&
+          (CreateOptions & FILE_DIRECTORY_FILE)) {
+        return STATUS_NOT_A_DIRECTORY;
+      }
+
       // FILE_FLAG_BACKUP_SEMANTICS is required for opening directory handles
       handle =
           CreateFile(filePath, genericDesiredAccess, ShareAccess,


Liryna:
Hi @Kerbox ,

Thank you for the contribution 👍 very appreciate for taking the time looking at the issue.

What I don't understand is that if the request is made with `FILE_DIRECTORY_FILE` then `DokanFileInfo->IsDirectory` should be set and we are going to be in this case https://github.com/dokan-dev/dokany/blob/9197ed333ea2a5e2217ac1690439e66ed668e1ab/samples/dokan_mirror/mirror.c#L308

Then the file will be open as directory https://github.com/dokan-dev/dokany/blob/9197ed333ea2a5e2217ac1690439e66ed668e1ab/samples/dokan_mirror/mirror.c#L331 (that's our issue).
Maybe we can add your check just before this createfile ?
```
if (!(fileAttr & FILE_ATTRIBUTE_DIRECTORY) &&
      !(CreateOptions & FILE_NON_DIRECTORY_FILE)) {
        (CreateOptions & FILE_DIRECTORY_FILE)) {
    DokanFileInfo->IsDirectory = TRUE;
      return STATUS_NOT_A_DIRECTORY; 
}
```

I say this because your fix currently broke a couple of the test:
https://ci.appveyor.com/project/Maxhy/dokany/build/1.0.1-797/job/dhsf5biox3vqmk6u#L821

Kerbox:
I got all tests to pass here (on win7 and win10), but appveyor claims that base\08.t fails now.

Liryna:
@Kerbox 👍 Thank You
I have rerun the build appveyor and test passed. I think the file was just not created in time before the delete.

You can rebase and clean your commit for the merge.

Thanks again for taking the time to look into this !

Kerbox:
Never rebased before, I hope I did that correctly!

Liryna:
Perfect rebase @Kerbox 👍 Thanks for your contribution !

Issue #366, opened 2016-10-10
BSOD after file open


ajaxovic:
# Environment
 * Windows version: Win10Pro x64
 * Processor architecture: Inter Core i7-6700HQ
 * Dokany version: 1.0.0
 * Library type (Dokany/FUSE): Dokany
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [ ] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description
Hello All! I'm referencing closed debug on winsshfs app [here](https://github.com/Foreveryone-cz/win-sshfs/issues/78#event-817492164). It was labeled as dokan problem. I think, issue #344 is not same as mine, because browsing with file explorer works. So, here is problem:

I installed dokany version [1.0.0](https://github.com/dokan-dev/dokany/releases/tag/v1.0.0) and winsshfs version [1.6.1.13 devel RC5](https://github.com/Foreveryone-cz/win-sshfs/releases/tag/1.6.1). Then, successfully mounted my server. With total commander, navigating to some file and press F4 for edit = BSOD. Here is minidump:

[101016-9390-01.zip](https://github.com/dokan-dev/dokany/files/519427/101016-9390-01.zip)

Thanks for your work and for your help!

# Logs
Please attach in separate files: mirror output, library logs and kernel [logs](https://github.com/dokan-dev/dokany/wiki/How-to-Debug-Dokan).
In case of BSOD, please attach minidump or dump analyze output.

Liryna:
Hi @ajaxovic ,

This seems to be the same that https://github.com/dokan-dev/dokany/issues/345 but you faced a BSOD.
When you F4, notepad open and it is at the save that it BSOD ?

ajaxovic:
No, without saving. I just press F4, empty notepad win appears and immediately BSOD. Without any action. **But I have some new update**: It works only on files that already been on server. I can create file, do multiple open/close, update content. But when I edit some other file, that I did not create through mount point, windows crash.

Liryna:
Humm can you just try to install development tools. (Uninstall, install again but click on options but and check to install additional files).
[...]

PR #371, opened 2016-10-19
Double check that the returned security descriptor is valid

Kerbox:
To prevent BSOD in situations like #366


Kerbox:
For issue #366

Liryna:
Thank you @Kerbox for the quick fix ! This is very appreciated !!! 🏆 

Release 1.0.1 planned end of this week or next one with it 😎

Issue #200, opened 2016-03-17
net share does not work since Windows Server 2012 R2


shino0729:
# Environment
 * Windows version: Windows Server 2012 R2
 * Processor architecture: x64
 * Dokany version: v1.0.0-RC2
 * Library type (Dokany/FUSE): Dokany
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description bellow
  - [x] If I updated Dokan driver (installer or .sys), I tried to restart the computer before my tests
  - [x] It can be reproduced on latest [release and pre-release](https://github.com/dokan-dev/dokany/releases)

# Description
Cannot access to the shared mount point made by the mirror sample.

    > mirror.exe /r c:\tmp /l m
    > net share <share_name>=m:\ /grant:Everyone,full

Then access to the shared directory (\\\\127.0.0.1\\\<share_name\>), return the error "device is not ready".
 
On Windows Server 2008 R2, it works fine.

js69:
Just to add my observation, as net sharing a Dokan drive never worked for me on Win 8 and WIn 10: tested it now on Win7-x86 and it works there.

Liryna:
This is the job of the network provider (dokannp1.dll).
You can find the source here: https://github.com/dokan-dev/dokany/blob/master/dokan_np/dokannp.c

It is a very simple implementation and it is probably now outdated for new systems after Win7 as you describe.

I think it need to be reviewed/rewrite.

js69:
Could we learn something from the VirtualBox implementation of a network provider?

http://www.virtualbox.org/svn/vbox/trunk/src/VBox/Additions/WINNT/SharedFolders/np/vboxmrxnp.cpp

Liryna:
@shino0729 do you have the possibility to change registry:
[HKEY_LOCAL_MACHINE\SYSTEM\ControlSet001\Control\NetworkProvider\HwOrder]
[HKEY_LOCAL_MACHINE\SYSTEM\ControlSet001\Control\NetworkProvider\Order]

Rename in ProviderOrder DokanNP1 to Dokan1 and make your test again ?

I am currently not able to use net share on my VM:
```
>net share test=m:\ /grant:MyUser,FULL
The operation is invalid on a redirected resource.
```
[...]

PR #373, opened 2016-10-21
Return correct status for FSCTL_FILESYSTEM_GET_STATISTICS

Kerbox:
Need to return STATUS_INVALID_DEVICE_REQUEST, as described [here](https://msdn.microsoft.com/en-us/library/ff469604.aspx). This is a fix for #200


Liryna:
Oh really ! @Kerbox ! I will make the test tomorrow ❤️ ❤️ ❤️ !!!

Amazing if it is so simple 👍 Great job !

Maxhy:
I missed that one when trying to fix it months ago...
After testing, it is indeed working. Thanks a lot, good job!

Liryna:
This very appreciate @Kerbox ! you can even imagine 😄 !

Thanks for taking a look at current open issue on dokan ❤️

Kerbox:
No problem! I am glad I can help. 😄

PR #377, opened 2016-10-26
Add Windows 2016 support to README

Rondom:
I have tested Dokany on the final version of Windows Server 2016 running in a qemu-VM with SecureBoot enabled.


Liryna:
Oh @Rondom ! Totally forgot it >< ! All this work and even not in the read me 😢 

Thank you 👍 !

PR #378, opened 2016-10-27
CI: AppVeyor speed-ups and enhancements

Rondom:
This set of commits fixes a few issues and decreases the time needed for a build.

Issues:
* command output was not appearing in log
* documentation push and Coverity Scan were run when using AppVeyor in one's private fork
* MSBuild output was not appearing in Messages tab in AppVeyor

The other part of the patches are some speed-ups. We cache downloaded packages to reduce load on the mirrors and decrease build-times. The individual speed-ups are not very big for some of the commits, because the cache is downloaded as a zip-archive at the beginning of the build as well and it does not matter which server you download from. There are some improvements underway from the AppVeyor side, so this might get faster in the future. Download-locations can be flaky though, so even without huge speed-ups  the cache should improve things. For example I have seen the Cygwin-Setup-download taking longer sometimes. Should any problems arise, can we move individual dirs out of the cache using environment variables at the top.
For a closer analysis, one can hover over the line numbers in the AppVeyor build-output to see when each line was printed. Though one needs to build at least one commit where `appveyor.yml` is not modified to see the cache in effect.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,7 @@
       DOKAN_CI_CACHE: C:\dokan_ci_cache
       CYG_CACHE: '%DOKAN_CI_CACHE%\cygwin'
       MSYS2_CACHE: '%DOKAN_CI_CACHE%\msys2'
+      CHOCO_CACHE: '%DOKAN_CI_CACHE%\choco'
 
   os: Visual Studio 2015
   version: 1.0.1-{build}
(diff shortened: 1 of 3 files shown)


Liryna:
Hi @Rondom !

That's perfect 😮 !
I didn't know cygwin cache could be merged for x86 and x64 ! Thats going to speed up yes ^^

Cache doxygen is also a good idea !

`test: off` I didn't know appveyor would look a test by default 😞  !

And specially thank you for adding RDP access by default that's awesome 😄 !

Keep rocking like that @Rondom ! very appreciate such contribution !
We gonna see the difference in build after the second build when he will have everything in cache.

PR #382, opened 2016-10-30
sys: Less wide locking

taruti:
This makes locking less wide with especial care to FsRtl locking routines.

Will do testing on monday.

### DIFF
--- a/sys/cleanup.c
+++ b/sys/cleanup.c
@@ -115,6 +115,7 @@ Return Value:
     RtlCopyMemory(eventContext->Operation.Cleanup.FileName,
                   fcb->FileName.Buffer, fcb->FileName.Length);
 
+    // FsRtlCheckOpLock is called with non-NULL completion routine - not blocking.
     status = FsRtlCheckOplock(DokanGetFcbOplock(fcb), Irp, eventContext,
                               DokanOplockComplete, DokanPrePostIrp);
     DokanFCBUnlock(fcb);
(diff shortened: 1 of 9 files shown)


taruti:
Copied+deleted 100k files and tried with mirror + one custom filesystem.

Liryna:
Thank you @taruti this seems to be an elegant way to do it 👍 

Just have a couple of questions pending and I merge it ❤️

taruti:
@Liryna feel free to ask any questions :)

Liryna:
@taruti They are in the comments of the PR 😃 
https://github.com/dokan-dev/dokany/pull/382/files

taruti:
For some weird reason Github is not displaying those comments on this PR to me.

Liryna:
Ok probably the new review system of github 😄 
-are you sure about this ? I agree that it is a struct full of ULONG. But should not be locked to be sure that all fields are updated in the same time ? LN340
- I guess this is linked to all locks removed in fscontrol.c . Did you find a documentation that say it is multi thread safe ? fastfat sample seems to lock FCB during FsRtlCheckLockForReadAccess https://github.com/Microsoft/Windows-driver-samples/blob/aa6e0b36eb932099fa4eb950a6f5e289a23b6d6e/filesys/fastfat/read.c#L1096 
LN 346

Liryna:
Seems like you are right, from osr thread, after FsRtl init the struct seems to be protected internally. 👍 
http://online.osr.com/showthread.cfm?link=243153

Thank you for your contribution @taruti as always!

Issue #386, opened 2016-11-02
BSOD during FSTest Config on CI


Rondom:
We cause BSOD on AppVeyor which is the reason for the restarts in the the middle of FSTest.

I made AppVeyor upload the dump as an Artifact. You can download a dump here: https://ci.appveyor.com/project/Rondom/dokany/build/1.0.1-216/artifacts

Liryna:
What O_o ? !
(dump is missing)

Rondom:
> (dump is missing)
Working on it ;-)

Rondom:
MEMORY.DMP can be found here: https://ci.appveyor.com/project/Rondom/dokany/build/1.0.1-216/artifacts
The build will still be active for an hour with RDP enabled.

Liryna:
can I connect to it ?

Rondom:
Yes

Liryna:
Ok this come from https://github.com/dokan-dev/dokany/commit/2768ac75402a0dbe06de163763485e251be2ac54

```
0: kd> !analyze -v
*******************************************************************************
*                                                                             *
*                        Bugcheck Analysis                                    *
*                                                                             *
*******************************************************************************

DRIVER_PAGE_FAULT_BEYOND_END_OF_ALLOCATION (d6)
N bytes of memory was allocated and more than N bytes are being referenced.
This cannot be protected by try-except.
When possible, the guilty driver's name (Unicode string) is printed on
the bugcheck screen and saved in KiBugCheckDriver.
Arguments:
Arg1: ffffcf8047847180, memory referenced
Arg2: 0000000000000000, value 0 = read operation, 1 = write operation
Arg3: fffff80026008874, if non-zero, the address which referenced memory.
Arg4: 0000000000000000, (reserved)

Debugging Details:
------------------


DUMP_CLASS: 1

DUMP_QUALIFIER: 401

BUILD_VERSION_STRING:  9600.18378.amd64fre.winblue_ltsb.160611-0600

SYSTEM_MANUFACTURER:  Microsoft Corporation

VIRTUAL_MACHINE:  HyperV

SYSTEM_PRODUCT_NAME:  Virtual Machine

SYSTEM_SKU:  None

SYSTEM_VERSION:  Hyper-V UEFI Release v1.0

BIOS_VENDOR:  Microsoft Corporation

BIOS_VERSION:  Hyper-V UEFI Release v1.0

BIOS_DATE:  11/26/2012

[...]

PR #389, opened 2016-11-03
AppVeyor Enhancements: BSOD, clang-format, error-checking

Rondom:
1. Check return code of external commands faa9485. We want to fail early if external commands fail. Thus we introduce a function that checks the $LASTERRORCODE and use it throughout the code. This makes failures much easier to spot because the failure is near the cause of the error. Motivation for this was a failing build where the [Cygwin-installation failed half-way-through due too a flaky mirror](https://ci.appveyor.com/project/Maxhy/dokany/build/1.0.1-823/job/f2t221q6qxrpvyyd#L196) and I spent quite some time wondering about the [why cmake was trying to use MSVC instead of GCC](https://ci.appveyor.com/project/Maxhy/dokany/build/1.0.1-823/job/f2t221q6qxrpvyyd#L4771.) Turned out that native Windows-cmake is in the path and thus got run instead.
2. Run `git clang-format` on pull-requests c8ffc49. We output an AppVeyor-message in case the code format on the lines changed does not match with the one defined in `.clang-format`. This  works only on pull requests and not on normal direct pushes to the repo, because AppVeyor does not provide the commitref *before* the push which is needed in order to determine the lines to check.
    
   - Demo 1: [PR with dirty indentation](https://github.com/Rondom/dokany/pull/3/files), Appveyor Build with [warning message](https://ci.appveyor.com/project/Rondom/dokany/build/1.0.1-241/job/85mh41qdx3wwrpu4/messages) and [suggested changes for download](https://ci.appveyor.com/api/buildjobs/85mh41qdx3wwrpu4/artifacts/clang_format_8ac0717_into_eca796c.diff)
   - Demo 2: [PR with clean indentation](https://github.com/Rondom/dokany/pull/4/files), Appveyor Build [with info message](https://ci.appveyor.com/project/Rondom/dokany/build/1.0.1-244/job/u00jtl7w4khq1niw/messages)

3. Reboot clean-up and BSOD-Detection c405310, 70b08f5: Motivation for this was issue #386. This is by no means perfect, because AppVeyor does not behave deterministically in base of a BSOD (see my comment here, but at least in some cases we are able to detect the BSOD and even upload the `MEMORY.DMP` as an artifact for offline debugging.

4. Some minor changes 9221bee, e26551f, eca796c

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -99,6 +99,24 @@
         }
 
   before_build:
+  - ps: |
+      # checking for BSOD
+      $bugchecks = (Get-EventLog -LogName system | Where-Object {$_.eventID -eq '1001' -and $_.Source -eq 'BugCheck'} | Select-Object -Property *)
+      if ($bugchecks) {
+        Write-Host "It seems like we rebooted due to a bugcheck (BSOD)."
+        Format-List -InputObject $bugchecks | Out-String | Out-Host
+        $memdumpArchive = "memory_dmp_$(& git rev-parse --short HEAD)_${env:APPVEYOR_BUILD_VERSION}.7z"
+        Write-Host "Compressing MEMORY.DMP"
+        Exec-External { &7z a -t7z $memdumpArchive "$env:WINDIR\MEMORY.DMP" }
+        Push-AppveyorArtifact $memdumpArchive
+        Write-Host "MEMORY.DMP uploaded as build artifact $memdumpArchive"
+        Add-AppveyorMessage `
+          -Category Error `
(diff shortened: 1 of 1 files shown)


Liryna:
Hi @Rondom ,

As always, big thank you 🏆 !!! You are really rocking it !
This features are awesome !
I always wanted a clang-format check 😄 !!! We will have beautiful pull request to come with it ❤️ I am sure.

BSOD is also very important as we saw ! A build could succeed even if it has BSOD 😢 Very good point here !
 
Very clean PR as usual ^^ !

Issue #391, opened 2016-11-05
Win7 - mirror.exe without administrative privileges, running executables


Nekotekina:
Hi, I'm experimenting with Dokan. If I run mirror.exe without administrator privileges, trying to run any simple executable from the mounted drive fails with the error 0xC000000D (STATUS_INVALID_PARAMETER). OS: Windows 7 SP1 x64. Executables I tried don't require any privileges.

The questions are (not strictly related to mirror.exe):
1) Why this happens and is it possible to modify mirror sample to fix it?
2) Is it generally possible to run unprivileged executables from virtual filesystems without NTFS features?
#132 seems very related but I failed to find exact answers.

Liryna:
@Nekotekina Hi, 

Can you look into the logs to see when this status is returned?

Nekotekina:
In Dokan logs, there is no such status. This must be process exit status, from Procmon:

19:05:58,5776677	Clockres.exe	11908	Process Start		SUCCESS	Parent PID: 2408
19:05:58,5776710	Clockres.exe	11908	Thread Create		SUCCESS	Thread ID: 10940
19:05:58,5916653	Clockres.exe	11908	Load Image	C:\Windows\System32\ntdll.dll	SUCCESS	Image Base: 0x76ec0000, Image Size: 0x1aa000
19:05:58,5919298	Clockres.exe	11908	Load Image	C:\Windows\SysWOW64\ntdll.dll	SUCCESS	Image Base: 0x770a0000, Image Size: 0x180000
19:05:58,5920283	Clockres.exe	11908	CreateFile	C:\Windows\Prefetch\CLOCKRES.EXE-7D8C1BA1.pf	NAME NOT FOUND	Desired Access: Generic Read, Disposition: Open, Options: Synchronous IO Non-Alert, Attributes: n/a, ShareMode: None, AllocationSize: n/a
19:05:58,5920566	Clockres.exe	11908	Thread Exit		SUCCESS	Thread ID: 10940, User Time: 0.0000000, Kernel Time: 0.0000000
19:05:58,5920710	Clockres.exe	11908	QueryNameInformationFile	C:\Windows\System32\apisetschema.dll	SUCCESS	Name: \Windows\System32\apisetschema.dll
19:05:58,5920876	Clockres.exe	11908	QueryNameInformationFile	\Device\Volume{d6cc17c5-1714-4085-bce7-964f1e9f5de9}\Clockres.exe	SUCCESS	Name: \Clockres.exe
[...]

PR #392, opened 2016-11-08
Mirror: Return empty SACL if mirror doesn't have SeSecurityPrivilege

Kerbox:
If mirror cannot get the SE_SECURITY_NAME privilege, don't ask for SACL in MirrorGetFileInformation. This will confuse win7 for some reason, as in #391.


Liryna:
Oh @Kerbox 😍 ! I had an idea the issue would be linked to this but did not have time to do it.

Thanks a lot 👍 ! I very appreciate your work @Kerbox and very well found !

PR #394, opened 2016-11-09
Align security descriptor to 4-byte boundary in DokanDispatchSetSecurity

Kerbox:
This is so that it can easily be used with win32 functions without having to copy the security descriptor into a separate aligned buffer.


Liryna:
Thank you @Kerbox for the fix ! As always, very appreciate 😋 !

PR #405, opened 2016-11-28
Fix BSOD. When drive ist started using n option and procmon is attaced

marinkobabic:
Networkdrive has no slash root folder. The diskdevicename must be returned in this case

### DIFF
--- a/sys/fileinfo.c
+++ b/sys/fileinfo.c
@@ -32,6 +32,7 @@ DokanDispatchQueryInformation(__in PDEVICE_OBJECT DeviceObject, __in PIRP Irp) {
   ULONG info = 0;
   ULONG eventLength;
   PEVENT_CONTEXT eventContext;
+  BOOLEAN isNormalized = FALSE;
 
   // PAGED_CODE();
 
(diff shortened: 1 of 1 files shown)


Liryna:
Hi @marinkobabic ,

I tested your fix it correctly fix the network behavior 👍 congratulations !
Unfortunately, when I try to copy some files (10-20) multiple times, I get a BSOD BAD POOL 😢 

My virtualbox does not seem to save the kernel dumps (for a mysterious reason) so I have unfortunately no call stack to give but this is 100% easy to reproduce.

marinkobabic:
Hi Liryna

You copied the files from your local drive to dokan drive?

Procmon was running in background?

Thx
Marinko

Liryna:
No, procmon was not running.

I just select a couple of files that are in the mirror and spam CTRL+C & CTRL+V a couple of times and I get a BSOD.

marinkobabic:
No way to reproduce it. What os are you using? My one is Windows 10 x64. I
copied files including folders several times but no BSOD. If you attach the
kernel debugger you should get more details. Can you reproduce the BSOD
without the pull request?

Rondom:
@marinkobabic Not relevant to the bug, but in its current form, your PR introduces some build warnings: https://ci.appveyor.com/project/Maxhy/dokany/build/1.0.1-25/job/qrc14ku9fxmhykdl

marinkobabic:
Will fix it. It's just a backslash 😆 thx

Liryna:
@marinkobabic I tried to reproduce without the PR and currently not have been able and when I moved to the PR again to do the test I am currently not able to reproduce it.
I think the BSOD probably come from another reason (copy lots of files seems to make it happen sometimes).

I will be happy to merge your PR when the warning (@Rondom 👍  ) will big fixed .

Rondom:
@marinkobabic The build-failure is not your fault, pull #407 to fix it or rebase on top of master once @Liryna merges it.

@marinkobabic It looks like FSTest hangs for some mysterious reason. I would guess it is a BSOD. If you cannot reproduce it locally, add your fork to AppVeyor (you can register/login using GitHub) and use [RDP](https://github.com/dokan-dev/dokany/blob/a8bfb27e7f33c8486c049b1f400f3ec6ea577afd/appveyor.yml#L27-L33) to help with debugging.

marinkobabic:
The question is, can it be reproduced on AppVayor? Will try also to
reproduce it locally.

marinkobabic:
Using this fix procmon displays m:\device\...mydir instead of m:\mydir which is not fine. So need more time to review the problem. Please do not merge it. Will cancel the pull request.

marinkobabic:
have enabled it only when network option is used. So we have no risk and all should work as usually. The BSOD should be gone.

Liryna:
I have no been able to reproduce the BSOD @marinkobabic 

Thank you for the fix and for taking the time ! very appreciated as always 👍

PR #407, opened 2016-11-30
CI: Temporary fix for bug in MSYS2-MINGW-CMake

Rondom:
The MSYS2-MINGW-CMake packaging (or one of its dependencies is missing a dependency
to libxml2) This breaks our build, see: https://ci.appveyor.com/project/Maxhy/dokany/build/1.0.1-27/job/ova2o24yyjh9i009#L4739.

How I troubleshooted this:
The error-message with libuv is a red herring. Nothing is wrong with libuv from what I can see and on my system stderr complains about zlib instead of libuv as it does on AppVeyor (both are installed). Running the executable from the GUI will produce the correct error message as a MessageBox.

This commit can be reverted once the bug is fixed:
Alexpux/MINGW-packages#1991

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -96,6 +96,8 @@
           bash "pacman -Syu --noconfirm --cache `"$unix_msys2_cache`""
           # install MinGW toolchain
           bash "pacman --sync --noconfirm --cache `"$unix_msys2_cache`" mingw-w64-{x86_64,i686}-toolchain mingw-w64-{x86_64,i686}-cmake"
+          # temporary workaround for bug https://github.com/Alexpux/MINGW-packages/issues/1991
+          bash "pacman -S --noconfirm --cache `"$unix_msys2_cache`" mingw-w64-{x86_64,i686}-libxml2"
         }
 
   before_build:


Liryna:
Thank you @Rondom ! very appreciate as always 👍 

We will need to remove this line when the issue is fixed on their side ?

Rondom:
It does not do any harm, when it's still there, but yes, we should revert this commitonce it is fixed

Rondom:
Now you merged it

Rondom:
Maybe I spoke too soon. Seems like they were very fast in fixing it.
Alexpux/MINGW-packages#1991

Now you merged it already. Can you test, if the fix has hit the mirror, yet? If rerunning an older build still works, we can revert that commit straight away. 😄

Liryna:
Done :) https://github.com/dokan-dev/dokany/pull/408
We will see the build result

Rondom:
No need to merge it prematurely. If the fix has not been propagated to the mirrors, you might need to apply it again... We will see ;-)

Liryna:
Too quick 😢 looks like
Lets wait that the mirrors get updated 😄

Issue #427, opened 2017-02-01
Where is the pdb file for dokanfuse?


netheril96:
# Environment
 * Windows version: 10
 * Processor architecture: x64 
 * Dokany version: 1.0.2
 * Library type (Dokany/FUSE): FUSE 
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the [latest version](https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
  - [ ] I tested one of the last [snapshot](https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot) from appveyor CI

# Description

The DokanSetupDbg.exe does not come with PDB files for fuse dlls. It makes debugging mysterious errors from Dokany itself much harder (I am encountering `std::system_error` during the `filler` call passed to `readdir`).

Rondom:
Indeed, no PDB-file is installed for the MSVC-compiled dokanfuse1.dll. Nor are there any debug-symbols inside the MingW- (libdokanfuse1.dll) or Cygwin-DLL (cygdokanfuse1.dll)...

For now, you can [build Dokan yourself](https://github.com/dokan-dev/dokany/wiki/Build).

I can prepare a PR for this and the other issue, but I will not have much time in the next days, so no guarantees...

Rondom:
I have created PR #429. Please test it and give feedback, because I cannot test it myself right now.

This will only fix the missing PDB for the MSVC-build. Debugging symbols for the GCC-build are still missing. I have decided to postpone this, because that will be better done as part of a some refactoring of the dokanfuse build.ps1

Liryna:
PR of @Rondom ( :heart: ) add the missing PDB for the MSVC-build fuse.
I let the ticket open until fuse cygwin debug info are also not embedded.

PR #429, opened 2017-02-03
WiX: Ship PDB-files for dokanfuse.dll

Rondom:
**This is completely untested. I have never tried installing, nor have I tried using the generated PDB to debug anything. Please double-check and test.**

Create debugging symbols both for Release and Debug configuration of the
MSVC dokanfuse1.dll and ship them in the installer.
See #427.


Liryna:
Hi @Rondom ,

Thank you for the contribution ❤️ I will try to test the installer locally this weekend.

Liryna:
Thank you @Rondom I made the test and the Dbg installer correctly install the fuse pdf. 🏆

PR #430, opened 2017-02-03
CI: Don't write clang-format diff if no changes in C-files

Rondom:
This fixes a small issue where the script did not account for the case where
no C-files where modified in a PR and git-clang-format wrote a
.diff-file with the text "no modified files to format".

[Example-PR where it occurred](https://ci.appveyor.com/project/Maxhy/dokany/build/1.0.1-62/job/v9r38yu1ciowoqm6/messages)

Ideally clang-format would print messages like this to stderr. As a
workaround we are grepping for the message to filter it out.


Liryna:
Hi @Rondom ,

Thank you for the contribution 👍 and well found 😃 !

PR #438, opened 2017-02-08
Fix use-after-free in GetDokanOpenInfo().

pliard-chromium:
The DOKAN_OPEN_INFO struct is ref-counted in userspace while an opaque pointer
to it is also stored, though not dereferenced, in kernel space via the
EVENT_CONTEXT struct.

Ref-counting is needed to manage the lifetime of DOKAN_OPEN_INFO as Dokan uses
loosely synchronized worker threads in userspace to dispatch the actual file
system requests.

When the ref-count of the DOKAN_OPEN_INFO goes to 0, the DOKAN_OPEN_INFO
pointer gets free'd. However worker threads other than the thread doing the
free() could still observe a stale copy of the pointer in their EVENT_CONTEXT
struct which each worker thread has a unique instance of.

This changes the user-space library so that there is an extra indirection to
lookup DOKAN_OPEN_INFO pointers. Instead of directly dereferencing a
potentially stale DOKAN_OPEN_INFO pointer, GetDokanOpenInfo() now looks up the
pointer in a table of entries indexed by a monotonically increasing ID which is
also stored in the EVENT_CONTEXT struct.

See issue 436 for more details.

### DIFF
--- a/dokan/create.c
+++ b/dokan/create.c
@@ -330,9 +330,11 @@ VOID DispatchCreate(HANDLE Handle, // This handle is not for a file. It is for
   if (origFileName)
     free(origFileName);
 
+  if (eventInfo.Status != STATUS_SUCCESS) {
+    free((PDOKAN_OPEN_INFO)(UINT_PTR)eventInfo.Context);
+    eventInfo.Context = 0;
+  }
+
   SendEventInformation(Handle, &eventInfo, sizeof(EVENT_INFORMATION),
                        DokanInstance);
-
-  if (eventInfo.Status != STATUS_SUCCESS)
-    free((PDOKAN_OPEN_INFO)(UINT_PTR)eventInfo.Context);
 }


marinkobabic:
Two different threads do  FindOpenInfoEntry and one of them is doing free. How does the new concept solve the problem that one threads has in this case a stale DOKAN_OPEN_INFO and it doesn't operate on it?

pliard-chromium:
GetDokanOpenInfo() does a thread-safe OpenCount increment so my understanding is that once a worker thread looked up a DOKAN_OPEN_INFO pointer by ID in the open info table, the pointer is guaranteed to remain valid until DokanReleaseOpenInfo() is called.
Are there still cases in the codebase where DOKAN_OPEN_INFO pointers are dereferenced without a prior OpenCount increment? If that's the case I missed that.

DOKAN_OPEN_INFO pointers are also now only free'd via DokanReleaseOpenInfo() which happens within the DokanInstance critical section like GetDokanOpenInfo() and FindOpenInfoEntry() do.

It's very possible that I'm missing something though. Would you mind elaborating if you still think there is a problem?

marinkobabic:
I'm still thinking about the real source of the problem. When you check the code before your changes, every DispatchCreate creates DOKAN_OPEN_INFO because the driver told the usermode to do so. In case of DispatchCleanup and DispatchClose the same. If the driver says that the file has been closed, so the DOKAN_OPEN_INFO can be destroyed for sure. So the question why is an OpenCount needed at all in the usermode library? SendEventInformation must not ReleaseDokanOpenInfo as well (this is the job of close). ClearFindData(openInfo->DirListHead) clears the list without to lock the access and so on. If the code is cleaned up, then your should not face the problem you originally had and usermode does not care about the OpenCount and when the DOKAN_OPEN_INFO should be destroyed.

This should be the starting point of our discussion. Maybe is my theory totally wrong but let's also see what other contributors say. I'm also very interested in your opinion.

pliard-chromium:
I have very limited experience with this codebase but it looks to me that this OpenCount is present in userspace mostly to allow Dokan in userspace to dispatch the requests to multiple worker threads instead of e.g. asynchronously dispatching the requests to a single thread which pull request #210 implements I believe. I agree with you, this feels a little wrong.
These worker threads pulling events from the driver in particular can get scheduled in arbitrary order by the kernel so my understanding is that events sent by the driver to userspace may be dispatched in userspace out-of-order.

The fact that events are dispatched out-of-order can cause issues if multiple events are simultaneously dispatched while having dependencies between each other, e.g. DispatchQueryInformation() and DispatchClose(). There doesn't seem to be the guarantee in the current state that the driver will send the close event only once all the events affecting the same file completed. Or is there actually such guarantee? If that's already the case then I agree with you the OpenCount shouldn't be necessary at all in userspace and there would potentially be a bug in the driver if this is not always happening. 

This pull request basically turns a crash due to a use-after-free into an error which is not ideal either and I wholeheartedly agreed with you here, there is likely a bigger design issue that needs to be fixed.
The very large pull request #210 should at least sanitize the ordering in userspace. The question of whether the driver is currently ensuring that close is sent only on completion of all other events for the same file remains though.

pliard-chromium:
FYI I will try to find out whether we receive DispatchClose() while there are still outstanding requests for the same file handle. According to the IRP_MJ_CLOSE documentation it shouldn't be the case.

pliard-chromium:
I'm glad you gave me some push back. I found out what's really going on here and the good news is that it should be a very small and simple fix. I will update my patch. I should have caught this earlier but the issue is basically due to the end of DispatchCreate() which frees the DOKAN_OPEN_INFO pointer in case of error after having sent it as non-null to the driver. This leads to the driver providing us again with the stale context pointer. The kernel seems to do so only for certain error codes. I'm seeing this in particular when my file system returns STATUS_TIMEOUT. Other error statuses seem to be fine.

I will see later in a follow-up pull request if I can also remove the OpenCount ref-counting. This should simplify things quite a bit.

pliard-chromium:
JFYI, this should be ready again for review.

pliard-chromium:
Also FYI: I changed my file system so that it returns STATUS_IO_TIMEOUT instead of STATUS_TIMEOUT during DispatchCreate() and the use-after-free also disappears that way. It seems that the kernel has some very particular handling for the STATUS_TIMEOUT error leading it not to fully discard the handle.

marinkobabic:
From my point of view it looks good. You should  squash the commits into one. Well found and fixed! :-)

pliard-chromium:
@marinkobabic Done, thanks!

Liryna:
Thank you for the PR @pliard-chromium , this also looks good for me !
Good work on this, it very appreciate ! ❤️

dokan/create.c
```diff
@@ -330,9 +330,11 @@ VOID DispatchCreate(HANDLE Handle, // This handle is not for a file. It is for
   if (origFileName)
     free(origFileName);
 
+  if (eventInfo.Status != STATUS_SUCCESS) {
```
pliard-chromium: FYI, I have been doing more digging regarding this STATUS_TIMEOUT vs STATUS_IO_TIMEOUT issue, regarding the fact in particular that the Context pointer gets sent back to user space with STATUS_TIMEOUT but doesn't with STATUS_IO_TIMEOUT.

I believe there is a discrepancy between the status check here and the one in the driver in DokanCompleteCreate(). The driver uses the NT_SUCCESS() macro which says "Evaluates to TRUE if the return value specified by Status is a success type (0 − 0x3FFFFFFF) or an informational type (0x40000000 − 0x7FFFFFFF)". STATUS_TIMEOUT happens to be in that range while STATUS_IO_TIMEOUT is an actual error.

I think we should probably make this check use NT_SUCCESS() instead. I'm worried that there might be other discrepancies in the codebase. What do you think?

marinkobabic:
There are maybe many other STATUS codes which can be used and are as well
in this range. So from my point of view you are absolutely right. Other
STATUS_SUCCESS checks should as well be verified.

Am 17.02.2017 00:25 schrieb "pliard-chromium" <notifications@github.com>:

Liryna:
@pliard-chromium I agree that we should add the NT_SUCCESS macro in the library (not present in user land) and use it instead of == STATUS_SUCCESS.
>#define NT_SUCCESS(Status) (((NTSTATUS)(Status)) >= 0)

pliard-chromium:
@Liryna @marinkobabic Thanks. I will send you a pull request using NT_SUCCESS() for that particular check. There might be other places in the codebase where it should be used as well in the future.

PR #440, opened 2017-02-09
Remove ref-counting of DOKAN_OPEN_INFO pointers.

pliard-chromium:
DispatchClose() is guaranteed to be called once all the outstanding
operations for the handle completed so ref-counting doesn't appear to be
necessary.

### DIFF
--- a/dokan/dokan.c
+++ b/dokan/dokan.c
@@ -483,10 +483,6 @@ VOID SendEventInformation(HANDLE Handle, PEVENT_INFORMATION EventInfo,
   ULONG returnedLength;
 
   // DbgPrint("###EventInfo->Context %X\n", EventInfo->Context);
-  if (DokanInstance != NULL) {
-    ReleaseDokanOpenInfo(EventInfo, DokanInstance);
-  }
-
   // send event info to driver
   status = DeviceIoControl(Handle,           // Handle to device
                            IOCTL_EVENT_INFO, // IO Control code
@@ -537,6 +533,7 @@ DispatchCommon(PEVENT_CONTEXT EventContext, ULONG SizeOfEventInfo,
   eventInfo->BufferLength = 0;
   eventInfo->SerialNumber = EventContext->SerialNumber;
 
+  DokanFileInfo->EventContext = EventContext;
   DokanFileInfo->ProcessId = EventContext->ProcessId;
   DokanFileInfo->DokanOptions = DokanInstance->DokanOptions;
   if (EventContext->FileFlags & DOKAN_DELETE_ON_CLOSE) {
@@ -555,7 +552,7 @@ DispatchCommon(PEVENT_CONTEXT EventContext, ULONG SizeOfEventInfo,
     DokanFileInfo->Nocache = 1;
   }
 
-  *DokanOpenInfo = GetDokanOpenInfo(EventContext, DokanInstance);
+  *DokanOpenInfo = (PDOKAN_OPEN_INFO)(UINT_PTR)EventContext->Context;
   if (*DokanOpenInfo == NULL) {
     DbgPrint("error openInfo is NULL\n");
     return eventInfo;
@@ -570,44 +567,20 @@ DispatchCommon(PEVENT_CONTEXT EventContext, ULONG SizeOfEventInfo,
   return eventInfo;
 }
 
-PDOKAN_OPEN_INFO
-GetDokanOpenInfo(PEVENT_CONTEXT EventContext, PDOKAN_INSTANCE DokanInstance) {
-  PDOKAN_OPEN_INFO openInfo;
-  EnterCriticalSection(&DokanInstance->CriticalSection);
-
-  openInfo = (PDOKAN_OPEN_INFO)(UINT_PTR)EventContext->Context;
-  if (openInfo != NULL) {
-    openInfo->OpenCount++;
-    openInfo->EventContext = EventContext;
-    openInfo->DokanInstance = DokanInstance;
-  }
-  LeaveCriticalSection(&DokanInstance->CriticalSection);
-  return openInfo;
-}
-
-VOID ReleaseDokanOpenInfo(PEVENT_INFORMATION EventInformation,
+VOID ReleaseDokanOpenInfo(PDOKAN_OPEN_INFO OpenInfo,
                           PDOKAN_INSTANCE DokanInstance) {
-  PDOKAN_OPEN_INFO openInfo;
   EnterCriticalSection(&DokanInstance->CriticalSection);
-
-  openInfo = (PDOKAN_OPEN_INFO)(UINT_PTR)EventInformation->Context;
-  if (openInfo != NULL) {
-    openInfo->OpenCount--;
-    if (openInfo->OpenCount < 1) {
-      if (openInfo->DirListHead != NULL) {
-        ClearFindData(openInfo->DirListHead);
-        free(openInfo->DirListHead);
-        openInfo->DirListHead = NULL;
-      }
-      if (openInfo->StreamListHead != NULL) {
-        ClearFindStreamData(openInfo->StreamListHead);
-        free(openInfo->StreamListHead);
-        openInfo->StreamListHead = NULL;
-      }
-      free(openInfo);
-      EventInformation->Context = 0;
-    }
+  if (OpenInfo->DirListHead != NULL) {
+    ClearFindData(OpenInfo->DirListHead);
+    free(OpenInfo->DirListHead);
+    OpenInfo->DirListHead = NULL;
   }
+  if (OpenInfo->StreamListHead != NULL) {
+    ClearFindStreamData(OpenInfo->StreamListHead);
+    free(OpenInfo->StreamListHead);
+    OpenInfo->StreamListHead = NULL;
+  }
+  free(OpenInfo);
   LeaveCriticalSection(&DokanInstance->CriticalSection);
 }
 
--- a/dokan/access.c
+++ b/dokan/access.c
@@ -38,7 +38,7 @@ HANDLE DOKANAPI DokanOpenRequestorToken(PDOKAN_FILE_INFO FileInfo) {
     return INVALID_HANDLE_VALUE;
   }
 
-  eventContext = openInfo->EventContext;
+  eventContext = FileInfo->EventContext;
   if (eventContext == NULL) {
     return INVALID_HANDLE_VALUE;
   }
--- a/dokan/close.c
+++ b/dokan/close.c
@@ -49,10 +49,8 @@ VOID DispatchClose(HANDLE Handle, PEVENT_CONTEXT EventContext,
   // SendEventInformation(Handle, eventInfo, length);
 
   if (openInfo != NULL) {
-    EnterCriticalSection(&DokanInstance->CriticalSection);
-    openInfo->OpenCount--;
-    LeaveCriticalSection(&DokanInstance->CriticalSection);
+    ReleaseDokanOpenInfo(openInfo, DokanInstance);
+    eventInfo->Context = 0;
   }
-  ReleaseDokanOpenInfo(eventInfo, DokanInstance);
   free(eventInfo);
 }
--- a/dokan/create.c
+++ b/dokan/create.c
@@ -122,6 +122,7 @@ VOID DispatchCreate(HANDLE Handle, // This handle is not for a file. It is for
   eventInfo.BufferLength = 0;
   eventInfo.SerialNumber = EventContext->SerialNumber;
 
+  fileInfo.EventContext = EventContext;
   fileInfo.ProcessId = EventContext->ProcessId;
   fileInfo.DokanOptions = DokanInstance->DokanOptions;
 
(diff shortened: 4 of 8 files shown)


Liryna:
Hi @pliard-chromium ,

This looks like a good idea and I agree with you that after close, normally there should not be any call again.
I will review it a little later to be sure on this 👍

dokan/dokan.c
```diff
@@ -570,44 +566,20 @@ DispatchCommon(PEVENT_CONTEXT EventContext, ULONG SizeOfEventInfo,
...
 
-PDOKAN_OPEN_INFO
-GetDokanOpenInfo(PEVENT_CONTEXT EventContext, PDOKAN_INSTANCE DokanInstance) {
-  PDOKAN_OPEN_INFO openInfo;
-  EnterCriticalSection(&DokanInstance->CriticalSection);
-
-  openInfo = (PDOKAN_OPEN_INFO)(UINT_PTR)EventContext->Context;
-  if (openInfo != NULL) {
-    openInfo->OpenCount++;
-    openInfo->EventContext = EventContext;
```
pliard-chromium: @Liryna, my patch is also removing this line in particular which I'm not sure is a good idea. Do you know why the per-thread EventContext pointer is stored in DOKAN_OPEN_INFO pointer while it is shared by multiple threads? This looks strange to me but the code might have been relying on this. I will see if this EventContext pointer can be moved out of the DOKAN_OPEN_INFO struct.

pliard-chromium: FYI, I just updated my patch so that the per thread EventContext pointer is no longer stored in the shared DOKAN_OPEN_INFO struct. Instead it's now stored in the DOKAN_FILE_INFO struct which is per thread if I understand correctly.

pliard-chromium:
I don't immediately see why codacy/pr is failing. ReleaseDokanOpenInfo() is still used in close.c as far as I can tell.

Liryna:
@pliard-chromium don't worry about it 😄 I had to tag ignore +80 warning like this.
I will look on the PR when will get more time sorry 😢

pliard-chromium:
@Liryna sounds good, thanks.

pliard-chromium:
@Liryna @marinkobabic FYI, I think this change might actually be unsafe to do in the current state. I have this change applied to my fork of Dokany and I'm seeing a use-after-free.

I think that issue 436 had multiple causes. Pull request 438 fixed the use-after-free in case an informational status was returned to Dokan. However I think it's still possible to trigger a use-after-free there. I'm seeing in particular that DispatchClose() gets called while there is still a DispatchDirectoryInformation() in flight.
I can only trigger this while using my Dokany-based file system with a high latency 3G network which makes me think that a timeout/IRP cancellation might be involved.

The Microsoft documentation for IRP_MJ_CLOSE says in particular that the close gets sent once "All outstanding I/O requests have been completed or canceled". I'm wondering in particular whether IRP cancellation happens synchronously or asynchronously?

Liryna:
@pliard-chromium humm normally as you have read IRP_MJ_CLOSE should be the last call 😢 Maybe yes there is a request to process but the kernel gave up (timeout, cancel,...) because userland took too much time ?
Therefore the userland is still using the reference that has been free in close ?

I think this sounds logic

pliard-chromium:
@Liryna That's my suspicion, yes. I think we should either fix the Dokan driver so that the close doesn't get sent while there are still pending cancelled operations, if that's possible at all, or keep the DOKAN_OPEN_INFO ref-count since it's already there.

Also note that the Dokan client's close callback is currently invoked in DispatchClose() as opposed to when the open info count drops to 0. This means that the client has to implement its own ref-counting for the opaque user context pointer stored in the DOKAN_FILE_INFO. Without changing this I also get a use-after-free in my client code in case of timeout. I will send you a separate pull request addressing that.

Should we delete this pull request?

marinkobabic:
The driver will do a close if nobody is referencing the file. How is it
possible that you get the use after free? Driver has the knowledge about
all the references. Please can you describe more detailed your specific
case or even write a simple test?

Am 02.03.2017 21:27 schrieb "pliard-chromium" <notifications@github.com>:

@Liryna <https://github.com/Liryna> That's my suspicion, yes. I think we
should either fix the Dokan driver so that the close doesn't get sent while
there are still pending cancelled operations, if that's possible at all, or
keep the DOKAN_OPEN_INFO ref-count since it's already there.

Also note that the Dokan client's close callback is currently invoked in
DispatchClose() as opposed to when the open info count drops to 0. This
means that the client has to implement its own ref-counting for the opaque
user context pointer stored in the DOKAN_FILE_INFO. Without changing this I
also get a use-after-free in my client code in case of timeout. I will send
you a separate pull request addressing that.

Should we delete this pull request?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/dokan-dev/dokany/pull/440#issuecomment-283770146>, or mute
the thread

pliard-chromium:
@marinkobabic I will try to reproduce this with the Dokan mirror by adding a sleep() in the DispatchDirectoryInformation callback. As I said above the issue happens when an operation times out.

pliard-chromium:
@marinkobabic @Liryna  I can successfully reproduce the use-after-free with the patch in this pull request (removing the DOKAN_OPEN_INFO ref-counting) and adding 'if (++g_counter % 10 == 0) { Sleep(30); }' at the very beginning of MirrorFindFiles().

I'm running the Dokan mirror with /r and /l and that's the stack trace that I get below after using Explorer for a few minutes:

>	dokan.dll!InsertTailList(_LIST_ENTRY * ListHead, _LIST_ENTRY * Entry) Line 91	C
 	dokan.dll!DokanFillFileDataEx(_WIN32_FIND_DATAW * FindData, _DOKAN_FILE_INFO * FileInfo, unsigned char InsertTail) Line 303	C
 	dokan.dll!DokanFillFileData(_WIN32_FIND_DATAW * FindData, _DOKAN_FILE_INFO * FileInfo) Line 312	C
 	mirror.exe!MirrorFindFiles(const wchar_t * FileName, int(*)(_WIN32_FIND_DATAW *, _DOKAN_FILE_INFO *) FillFindData, _DOKAN_FILE_INFO * DokanFileInfo) Line 743	C
 	dokan.dll!DispatchDirectoryInformation(void * Handle, _EVENT_CONTEXT * EventContext, _DOKAN_INSTANCE * DokanInstance) Line 554	C
 	dokan.dll!DokanLoop(_DOKAN_INSTANCE * DokanInstance) Line 435	C
 	dokan.dll!invoke_thread_procedure(unsigned int(*)(void *) procedure, void * const context) Line 92	C++
 	dokan.dll!thread_start<unsigned int (__cdecl*)(void * __ptr64)>(void * const parameter) Line 115	C++

marinkobabic:
We need an idea how this can happen.

Maybe this is the case:

Driver has multiple requests for a specific file. Those requests have been
forwarded to usermode and the os got the response STATUS_PENDING. So and
now at some point the dokan driver decides that some of this requests are
timed out and those requests are canceled. Usermode has no idea about it.
The os decides now to close the handle of the file and dokan driver
forwards this request to usemode. So there may now be multiple threads
running at the same time in usermode using exactly the same context. The
DispatchClose will free the Context which may be used by other threads.
This causes the problem.

This may be the reason why the usermode reference count has been
introduced. Create should set it to 1 and when completed not decrease it.
All other Dispatch operations should increase it by 1 when dispatched and
decrease it by 1when completed. The DispatchClose should just decrease it
by 1 (no increase). When the counter reaches 0 then the context should be
freed. The free may happen in any dispatch operation. DispatchClose can
occur before other operations are completed. Following this approach all
should work fine. The drawback is that there is a usermode reference count.
The name of the property should be ContextReferenceCount.
Maybe even your previous issue should be gone and no extra check for
invalid context is needed because the issue can happen in any operation.

This problem can't be solved inside of the driver. Drive has no idea what
Usermode is doing.

This is actually just an assumption why ReferenceCount exists and your
original problem can occur in any operation. This assumption has now to be
confirmed.

Thx Marinko

Am 02.03.2017 21:48 schrieb "pliard-chromium" <notifications@github.com>:

pliard-chromium:
@marinkobabic Thanks. Your explanation matches my current understanding. We seem to agree that the current DOKAN_OPEN_INFO ref-count is necessary, in the current state of things at least.

This brings me back to the note I made in an earlier comment: "Also note that the Dokan client's close callback is currently invoked in DispatchClose() as opposed to when the open info count drops to 0. This means that the client has to implement its own ref-counting for the opaque user context pointer stored in the DOKAN_FILE_INFO. Without changing this I also get a use-after-free in my client code in case of timeout. I will send you a separate pull request addressing that."

I have a patch ready to fix that problem by calling the client's close operation from ReleaseDokanOpenInfo() when the open info count drops to 0. Does that make sense to you? I can send a separate pull request for that and we can continue the discussion there.

Liryna:
I am not sure calling close in ReleaseDokanOpenInfo would be very great. This sounds like a hack but that's maybe the way to do ? No close on IRP close and call when release has to be free ?
What do you think @marinkobabic does this idea sounds bad ? Am afraid that the close will be called too late and we would face a sharing violation if a createfile follows the irp close that we still didn't proceed.

marinkobabic:
It doesn't feel correct. The code can be improved anyway. DOKAN_OPEN_INFO
can get a new flag IsClosed. This flag would be set if the driver has
called the DispatchClose. In SendEventInformation this flag can be checked
and if set the DeviceIoControl should not be called. This way the usermode
is aware about the actual state of the file inside of the driver.

You remember the *SendWriteRequest*? Also there the flag should be checked
and this way we can prevent unnecessary call to the driver. LastError can
be set to ERROR_OPERATION_ABORTED

This way the code would be clean and we would have the full transparency.

pliard-chromium:
@Liryna Any chance you could elaborate on the sharing violation issue? The change I had in mind was only deferring the invocation of the client's close callback until the open info count drops to 0. AFAICT this shouldn't affect Dokan itself given that it doesn't use the return value of that callback. Or am I missing something?

Liryna:
@pliard-chromium If the stream is not closed in close, it is possible that there is a race condition depending on the sharing mode that he was open.
I mean that the kernel can probably just after the close (since for him everything is clean normally after close) request a CreateFile that will create a sharing violation since the not closed stream has no sharing access. This error would be "unexpected" from him and probably display an error to the user.

That's why currently we open during write/read after the close with sharing.

Maybe what I explain will never happen but it is a possibility 😢

pliard-chromium:
@Liryna Thanks for the explanation, that makes sense to me. The sharing violation would be triggered by the logic that the Dokan client has to implement, e.g. in Dokan mirror, is that correct? We don't currently implement such logic in our Dokan-based file system although we definitely should.

Would the issue still persist if the sharing violation detection logic was implemented in the driver instead? Would that make sense at all?

Liryna:
@pliard-chromium Dokan kernel does implement sharing violation on kernel side. The issue that would appear here is that the kernel get a createfile, for him there is no sharing violation because the close has happened for him and does not expect that the user has kept a handle open on the file. Therefore the request pass the kernel goes to the FS user and the sharing violation happen because handle is possibly still opened.
Windows request the handle to be cleaned, and it should be like that or unexpected behavior like describe would happen 😢

sefai:
Hello there, I had a similar problem. I am not sure if it's the same use after free issue you are talking about here. Let me describe:

In the mirror sample, a file HANDLE is stored in DokanFileInfo->Context. In MirrorCleanup callback file is closed using CloseHandle with DokanFileInfo->Context(file HANDLE) and DokanFileInfo->Context is set to 0. In MirrorCloseFile callback DokanFileInfo->Context is being checked and if it is not 0, CloseHandle is called again. CloseHandle only throws an exception in Debug mode if an invalid file handle is passed, but does ignore it in Release mode, so there appears no problem with the mirror sample when this handle is closed in MirrorCleanup but due to use after free problem used again in MirrorCloseFile. 

But in my user mode application I am storing a heap allocated pointer to a custom struct in DokanFileInfo->Context. This is freed in Cleanup callback but somehow this pointer appears again in CloseFile callback and it was giving me a dangling pointer issue. So, I implemented my own control mechanism in my application to overcome this dangling pointer issue. 

I would be happy if this is resolved in driver, because I had to implement a three-tier lookup table and a locking mechanism. In each operation I am validating the context against the lookup table, lock the entry, and release the entry after the operation is done. This is introducing a lot of code complexity and some delay because of locks.

Liryna:
Hi @sefai ,

Does in DokanFileInfo->Context you set the context to null just after free it ?
If yes, you should have a context null in closefile and would have no issue.

Is it possible in your case that the custom struct is shared with two DokanFileInfo ? (two that point the same file but opened twice ? (CreateFile called 2 times)

sefai:
Yes @Liryna , I set it to null. As I said this is not specific to my struct and I am creating a new struct for each CreateFile callback. 

Same thing happens in mirror. But CloseHandle absorbs the error in Release mode and in Debug mode you can continue execution by ignoring the exception. I almost always get an exception in Debug mode while building a heavy project in mirror drive.

timofonic:
@pliard-chromium @Liryna @marinkobabic @sefai 

Any news about this? Any reason this is still open? Can someone reply about this? Thanks in advance :)

Liryna:
Merged with https://github.com/dokan-dev/dokany/commit/d3e3fcd9624cd1e9ed71dfd44edf28e6f590a14d

Issue #442, opened 2017-02-13
When did dokan free the MDL when read operation perform?


yin19941005:
# Environment
 * Windows version:  All Windows
 * Processor architecture: amd64
 * Dokany version: 1.0.2 release
 * Library type (Dokany/FUSE): Dokany
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the [latest version](https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
  - [ ] I tested one of the last [snapshot](https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot) from appveyor CI

# Description
I checked the call hierarchy for the function [`DokanAllocateMdl`](https://github.com/dokan-dev/dokany/blob/master/sys/dokan.c#L555) and [`DokanFreeMdl`](https://github.com/dokan-dev/dokany/blob/master/sys/dokan.c#L576), and I found that the read operation didn't call the free memory descriptor list (MDL) after it allocate the MDL.

<br>

From MSDN Document : 
> If a driver allocates an MDL to describe a buffer, it must explicitly release the MDL when operations on the buffer are done. [(https://msdn.microsoft.com/en-us/library/windows/hardware/ff549126%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396)](https://msdn.microsoft.com/en-us/library/windows/hardware/ff549126%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396)

<br>

The `DokanQueryDirectory` will allocate the MDL on [`/sys/directory.c#Line145`](https://github.com/dokan-dev/dokany/blob/master/sys/directory.c#L145) and free it on `DokanCompleteDirectoryControl` [`/sys/directory.c#Line371`](https://github.com/dokan-dev/dokany/blob/master/sys/directory.c#L371).

The `DokanDispatchQuerySecurity` will allocate the MDL on [`/sys/security.c#Line108`](https://github.com/dokan-dev/dokany/blob/master/sys/security.c#L108) and free it on `DokanCompleteQuerySecurity` [`/sys/security.c#Line187`](https://github.com/dokan-dev/dokany/blob/master/sys/security.c#L187).

[...]

PR #448, opened 2017-02-15
Add DokanFreeMdl for read operation

yin19941005:
Adding DokanFreeMdl mentioned in #442. Should fix the forgotten free.


Liryna:
@yin19941005 there is an empty merge commit in the history 😢 
maybe you will need to create another branch with the changes here to make it disappear

yin19941005:
Oh! Sorry for that! I am pretty new on using Git.
I will try to create another branch to make it disappear.

PR #449, opened 2017-02-15
Add DokanFreeMdl for read operation

yin19941005:
Forgotten free MDL in DokanCompleteRead. Add the DokanFreeMdl to fix. Fix for #442 .

I think this time should be correct. Please correct me if I messed up again.


Liryna:
Thank you @yin19941005 :+1: very appreciate for taking the time on this !

PR #450, opened 2017-02-17
Check event info status in DispatchCreate() with NT_SUCCESS() as in t…

pliard-chromium:
…he driver.

DokanCompleteCreate() in the driver uses the NT_SUCCESS() macro to check
whether it should free the CCB, FCB... while DispatchCreate() in
userspace was freeing the DOKAN_OPEN_INFO pointer when the event info
status was not equal to STATUS_SUCCESS.

This was leading to operations being sent to userspace with a null
DOKAN_OPEN_INFO pointer after DispatchCreate() returned a successful
non-STATUS_SUCCESS status such as STATUS_TIMEOUT.

### DIFF
--- a/dokan/create.c
+++ b/dokan/create.c
@@ -330,7 +330,7 @@ VOID DispatchCreate(HANDLE Handle, // This handle is not for a file. It is for
   if (origFileName)
     free(origFileName);
 
-  if (eventInfo.Status != STATUS_SUCCESS) {
+  if (!NT_SUCCESS(eventInfo.Status)) {
     free((PDOKAN_OPEN_INFO)(UINT_PTR)eventInfo.Context);
     eventInfo.Context = 0;
   }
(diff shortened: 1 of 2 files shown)


Liryna:
Thank you @pliard-chromium for the PR !
I am just wondering if we should not also use the macro for the operation ? Does it would add something ?

pliard-chromium:
Thanks for taking a look @Liryna. I haven't got a chance to look into this yet. Will let you know when it's ready.

Liryna:
Hi @pliard-chromium ,

Probably the macro should also be used here:
https://github.com/dokan-dev/dokany/blob/1cef44a52273c41bc17db2860d85082392d1e575/dokan/create.c#L89

Liryna:
Thank you @pliard-chromium for the PR 👍 if you find any other `status != STATUS_SUCCESS` that should be replaced, please do not hesitate !

pliard-chromium:
Thanks @Liryna! Sorry I didn't get a chance to address your earlier comment.

PR #451, opened 2017-02-18
CI: MSYS2 fixes and improvements

Rondom:
This fixes two issues with on AppVeyor MSYS2. See the commit-msgs for detailed descriptions.
1. Build was failing due to the way we upgrade MSYS2
2. Pacman was unnecessarily re-installing already installed packages.


Liryna:
👍 This msys 😄 

Thanks a lot @Rondom ! very appreciate !

Rondom:
@Liryna maybe restart the builds of the two outstanding PRs

Liryna:
👍 it is staring

Issue #436, opened 2017-02-08
When create the same name folder after moved it, the copy file operation failed


noairman:
# Environment
 * Windows version: 10
 * Processor architecture: 64 bit
 * Dokany version: 1.0.2.1000
 * Library type (Dokany/FUSE): dokany/mirror
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the [latest version](https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
  - [x] I tested one of the last [snapshot](https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot) from appveyor CI

# Description
Similar to the issue #422 ,#428 ,but not just the exe type file( txt,log,dll,zip,exe.pdf,deb file type)
![error](https://cloud.githubusercontent.com/assets/10110303/22719658/76e0ee84-ede1-11e6-865a-dfc7f1ba9f6c.jpg)

This is how i reproduce the issue:
 1.Create the new folder,named "1",at the root directory;
 2.copy the files "1.txt" in it
 3.move the folder "1" to other place;
 4.create the folder "1" again at the root directory;
5.copy the same file "1.txt" into the "1"

Similar to the issue #422 ,#428 
i will get that error "Could not find this item".
And the FileCount remains high with the STATUS_OBJECT_NAME_COLLISION .But this time it's not only the exe file ,but also other type file( txt,log,dll,zip,exe.pdf,deb file type).But the doc,ini file type is OK.

**This issue only can be reproduced after the move folder operation,but not the delete folder and the cut folder operation**.Maybe the IRP_MJ_CLOSE is not done for the file when it's parent folder is moved.

# Logs
[dokan.log.txt](https://github.com/dokan-dev/dokany/files/759489/dokan.log.txt)

I just tested again with snapshot https://ci.appveyor.com/project/Maxhy/dokany/build/job/0o4m6la7kebpn1w5/artifacts
and got the same issue

marinkobabic:
So let's explain what here happens.

[...]

PR #454, opened 2017-02-20
Flush any opened files before doing a rename

marinkobabic:
Fix here described issue https://github.com/dokan-dev/dokany/issues/436


Liryna:
Hi @marinkobabic ,

Thanks a lot for the PR and taking the time on this issue 👍 as always ! 🍺

Liryna:
@marinkobabic performance question, can we not just flush file in the directory ?

marinkobabic:
@Liryna That would make sense. I flushed all of them just because it was simpler to implement.

Liryna:
Ok I see @marinkobabic do you plan to make a patch for it ? Or I create an issue to remind ?

marinkobabic:
@Liryna 
Here you have the refactored code. Please do a carefully review of the code https://github.com/dokan-dev/dokany/pull/462

Issue #447, opened 2017-02-15
Mirror.exe crash when canceling the copy operation on Windows 7 accessing the shared dokan drive


yin19941005:
# Environment
The Windows Server installed dokan and share the dokan drive to the client. 

 * Windows version:  Windows 7 as client, Windows Server 2016 as Server with dokan installed
 * Processor architecture: x86-64
 * Dokany version: 1.0.2.1000 release
 * Library type (Dokany/FUSE): Dokany

# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the [latest version](https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
  - [ ] I tested one of the last [snapshot](https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot) from appveyor CI

# Set Up Description
I had un-commented the `MirrorDokanGetDiskFreeSpace` in mirror and changed the available size as 50 GB with total size as 100 GB.
```
static NTSTATUS DOKAN_CALLBACK MirrorDokanGetDiskFreeSpace(
    PULONGLONG FreeBytesAvailable, PULONGLONG TotalNumberOfBytes,
    PULONGLONG TotalNumberOfFreeBytes, PDOKAN_FILE_INFO DokanFileInfo) {
  UNREFERENCED_PARAMETER(DokanFileInfo);

  *FreeBytesAvailable = (ULONGLONG)53687091200;			// = 512 * 1024 * 1024 * 100
  *TotalNumberOfBytes = (ULONGLONG)107374182400;		// = 1024 * 1024 * 1024 * 100
  *TotalNumberOfFreeBytes = (ULONGLONG) 53687091200;	// = 512 * 1024 * 1024 * 100

  return STATUS_SUCCESS;
}
```

Added more DbgPrint on `/dokan/write.c` on [Line 40](https://github.com/dokan-dev/dokany/blob/master/dokan/write.c#L40) : 
```
DbgPrint("SendWriteRequest : status = %d, EventLength = %lu, BufferLength = %lu, returnedLength = %lu \n", status, EventLength, BufferLength, returnedLength);

  if (returnedLength == 0) {
	  DWORD errorCode = GetLastError();
[...]

PR #457, opened 2017-02-22
Fix crash issue cause by canceling the copy operation

yin19941005:
The root cause is because the dokan.sys will still return status_success when the corresponding IRP is not found cause by the user cancellation. With this fix, the driver will return status_cancelled and the dokan usermode dll will handle the status returned correctly. This fix will fix the issue #447 .


yin19941005:
Sorry, I found a typo on it. I will close this pull request and reopen a new one.

PR #458, opened 2017-02-22
Fix crash issue cause by canceling the copy operation

yin19941005:
The root cause is because the dokan.sys will still return status_success when the corresponding IRP is not found cause by the user cancellation. With this fix, the driver will return status_cancelled and the dokan usermode dll will handle the status returned correctly. This fix will fix the issue #447 .

Sorry for the typo in the previous pull request.


Liryna:
Thanks a lot for the PR and taking the time on this issue @yin19941005 🥇 
very appreciate !

Issue #369, opened 2016-10-12
Git commit crashing with mirror.exe


dimov-cz:
# Environment
 * Windows version: 10x
 * Processor architecture: x64
 * Dokany version: 1.0.0.5
 * Library type (Dokany/FUSE): Dokany
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description
I'm little bit dissappointed from 1.0.0.5, previous version(RC4) was working better for me. A lot of people now reports me in WinSshFS about not working git. 
I can confirm this with reproduction script for mirror.exe. 

**Before running this test i must warn you, you will very probably get explorer.exe crash, you will be unable to logout, open start menu, processes can go to some uninterruptible state and hardreset will be only option probably.**

1. Create mirror: `mirror.exe /r mirrored /l t /t 1 /d /s`
2. git clone some small repo and try to commit.
I use this test script: [wfsGitTest1.zip](https://github.com/dokan-dev/dokany/files/525713/wfsGitTest1.zip) 

If you restart mirror for second try u are unable to remove directory over mirror.
Maybe it's related to something already fixed, but not released. This one? https://github.com/dokan-dev/dokany/issues/345

Liryna:
Hi @dimov-cz ,

Yes the issue is related to #345 . I tested your script on a snapshot and it work 👍 
```
liryna@DESKTOP-T9B1IFA MINGW64 ~/Desktop
$ ./wfsGitTest1.sh
Cloning into 'emsdk.tmp'...
remote: Counting objects: 825, done.
remote: Compressing objects: 100% (8/8), done.
remote: Total 825 (delta 2), reused 0 (delta 0), pack-reused 817
Receiving objects: 100% (825/825), 393.60 KiB | 0 bytes/s, done.
Resolving deltas: 100% (517/517), done.
warning: LF will be replaced by CRLF in newfile.dat.
The file will have its original line endings in your working directory.
[master 0bcd92e] add new file
 1 file changed, 1 insertion(+)
 create mode 100644 newfile.dat
0
```

[...]

Issue #264, opened 2016-06-16
Visual Studio cannot create browsing database file when opening project on mirrored drive


bailey27:
# Environment
- Windows version: Windows 10 64-bit
- Processor architecture:  Intel Haswell (i5-4200U)
- Dokany version: 1.0.0-RC3
- Library type (Dokany/FUSE): Dokany
# Check List
- [x ] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [x ] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
- [x ] I can always reproduce the issue with the provided description below.
- [ x] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
# Description

It seems that Visual Studio cannot create the browsing database on a mirrored drive.

Steps to reproduce

c:
cd \
mkdir c:\tmp\mirror 
cd tmp\mirror
git clone https://github.com/bailey27/cppcryptfs.git

[set up for Dokany debugging and reboot]

[ cd to C:\Program Files\Dokan\Dokan Library-1.0.0\sample\mirror]

[run as administrator]

.\mirror.exe /t 1 /d /s /r c:\tmp\mirror /l m 2> c:\tmp\mirror_vs.log

Open M:\cppcryptfs\cppcryptfs.sln in Visual Studio 2015 Community Edition

Visual Studio displays a dialog that says 

"An error occurred while creating or opening the C++ browsing database file M:\cppcryptfs\cppcryptfs.sdf. IntelliSense and browsing information will not be available for C++ projects.

Ensure that Microsoft SQL Server Compact 4.0 is installed and that no other applications are accessing the file.  If this problem persists, delete this file and reopen the solution."

I looked in the mirror log, and I couldn't figure out what VS tried to do that failed that caused this to happen. 

If I try to create a new project on the mirrored drive, the same error happens.
# Logs

Here are the logs.  I'm not sure what the "library logs" are.  If it's logging from the dll, I think there weren't any instructions for obtaining them.  I seem to have had only kernel driver output in dbgview.  I logged it to a file. I logged mirror to stderr and redirected to a file.

Here are the logs in this zip file.

[dokan_vs_logs.zip](https://github.com/dokan-dev/dokany/files/318023/dokan_vs_logs.zip)

Liryna:
Hi @bailey27 ,

[...]

PR #462, opened 2017-02-24
Use filename check before doing flush. Remove code redudancy

marinkobabic:
The check of the filename is now implemented. It's suboptimal because the backslash should be added to the folder before the check is performed. This is not possible without allocation of additional memory. Anyway the changes improve the code a lot and code redundancy is eliminated.

### DIFF
--- a/sys/cleanup.c
+++ b/sys/cleanup.c
@@ -86,22 +86,7 @@ Return Value:
     fcb = ccb->Fcb;
     ASSERT(fcb != NULL);
 
-    if (fileObject->SectionObjectPointer != NULL) {
-
-      if (fileObject->SectionObjectPointer->ImageSectionObject != NULL) {
-        MmFlushImageSection(&fcb->SectionObjectPointers, MmFlushForWrite);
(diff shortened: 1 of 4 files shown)


Liryna:
Thank you @marinkobabic that's perfect !
I will review it when I can ❤️  !

Liryna:
Should we not also check if `fcb->SectionObjectPointers` is null in FlushFcb ?

This seems to be good 👍 for me

marinkobabic:
Is SectionObjectPointers a pointer or a struct? Can it be null? If it can
be null so we should check it. Have no access to my machine. I'm on
vacation for a week:-)

Am 28.02.2017 12:57 schrieb "Liryna" <notifications@github.com>:

Liryna:
My bad @marinkobabic ! SectionObjectPointers is a struct :) I failed you were right to check ImageSectionObject that is a pointer 👍 
Thanks again for the fix !

Will test https://github.com/dokan-dev/dokany/issues/369 and https://github.com/dokan-dev/dokany/issues/264 when I get free time (also in vacation 😄 Hope you have a great week !)

PR #468, opened 2017-03-02
CI: Try rebooting two times and reduce timeout

Rondom:
Turns out that waiting for 60mins was not so smart... 😊 

Using the changes below, builds should fail early and we can find out, whether the build-VM does not shut down or whether it does not come back after reboot.

Instead of waiting for an hour to reboot, try rebooting in two different
ways with a timeout of 4 minutes each. If the system has not rebooted
after those two tries, fail the build.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -40,7 +40,13 @@
           Write-Host "Before reboot"
           Start-Sleep -s 2
           Restart-Computer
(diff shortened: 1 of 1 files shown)


Liryna:
👍 @Rondom thats a good idea 😄 
I hope it the force reboot will fix some build error!
Thank you @Rondom as always !

PR #473, opened 2017-03-12
FUSE: Check for non-empty directories on delete

### LINKED ISSUE #270: FUSE - Check if Directory is empty before calling rmdir
Dokan FUSE should list directory fils before try to delete it.
If not empty, rmdir should not be called and STATUS_DIRECTORY_NOT_EMPTY should be returned.

Must be done here:
https://github.com/dokan-dev/dokany/blob/1b5263f04a5c301e82fc3dbeba01cf0e40d187a2/dokan_fuse/src/fusemain.cpp#L171

Rondom:
Fixes #270.

Two remaining questions before merging. The first one is more out of interest, the second one is more relevant to this PR.
1. @Liryna As you can see I put the check earlier than you suggested in #270, which is in line with [what I read about file-deletion on Windows](http://download.microsoft.com/download/4/3/8/43889780-8d45-4b2e-9d3a-c696a890309f/File%20System%20Behavior%20Overview.pdf). This is a general question about Windows: What happens when the deletion fails during cleanup? Is it true that there is no possibility to signal user space that a deletion failed?
2. Is there the possibility of race conditions?
    * FileSetDisposition is called
    * the DeleteDirectory callback enumerates the direntries, finds none
    * a file is placed into the directory
    * DeleteDirectory returns with no-error (because it was empty), signals that the directory is going to be deleted

### DIFF
--- a/dokan_fuse/src/fusemain.cpp
+++ b/dokan_fuse/src/fusemain.cpp
@@ -424,15 +424,66 @@ int impl_fuse_context::create_directory(LPCWSTR file_name,
   return ops_.mkdir(fname.c_str(), dirmask_);
 }
 
+int impl_fuse_context::readdir_filler_set_has_files(void *buf, const char *name,
+                                      const struct FUSE_STAT *stbuf,
+                                      FUSE_OFF_T off) {
+  bool *has_files = static_cast<bool *>(buf);
+  int ret = 0;
+  if (strcmp(name, ".") && strcmp(name, "..")) {
+    *has_files = true;
+    ret = 1;
+  }
+  return ret;
+}
+
+int impl_fuse_context::getdir_filler_set_has_files(fuse_dirh_t hndl, const char *name,
+                                             int type, ino_t ino) {
+    bool *has_files = reinterpret_cast<bool *>(hndl);
+    int ret = 0;
+    if (readdir_filler_set_has_files(has_files, name, nullptr, 0))
+      ret = -ECANCELED;
+    return ret;
+}
+
+
 int impl_fuse_context::delete_directory(LPCWSTR file_name,
                                         PDOKAN_FILE_INFO dokan_file_info) {
   std::string fname = unixify(wchar_to_utf8_cstr(file_name));
-
-  if (!ops_.getattr)
+  if (!ops_.getattr || !ops_.rmdir || (!ops_.readdir && !ops_.getdir))
     return -EINVAL;
 
   struct FUSE_STAT stbuf = {0};
-  return ops_.getattr(fname.c_str(), &stbuf);
+  int ret = ops_.getattr(fname.c_str(), &stbuf);
+
+  /* TODO: Should we check if the parent-dir is writable and return -EACCESS?
+   * (by using ops_.access or alternatively other means such as getattr)
+   */
+  if (ret < 0)
+    return ret;
+
+  bool has_files = false;
+  if (ops_.readdir) {
+    impl_file_handle *hndl = reinterpret_cast<impl_file_handle *>(dokan_file_info->Context);
+    fuse_file_info *p_finfo = nullptr;
+    if (hndl != nullptr) {
+      fuse_file_info finfo(hndl->make_finfo());
+      p_finfo = &finfo;
+    }
+    ret = ops_.readdir(fname.c_str(), &has_files, &readdir_filler_set_has_files, 0, p_finfo);
+  } else {
+    ret = ops_.getdir(fname.c_str(), reinterpret_cast<fuse_dirh_t>(&has_files), &getdir_filler_set_has_files);
+    /* We only propagate the return value of getdir() when has_files is set.
+     * To avoid reading the whole directory, we return an error in our callback
+     * and set has_files after we encountered the first regular directory entry.
+     * It is not exactly clear what getdir() is supposed to return in this case,
+     * but given that we return -ENOTEMPTY anyways, we do not need to care.
+     */
+  }
+  if (has_files) {
+    ret = -ENOTEMPTY;
+  }
+
+  return ret;
 }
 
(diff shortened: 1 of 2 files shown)


Liryna:
Hi @Rondom ,

Thank you for the PR ! this fix is very appreciate 👍 !

1. About deletion fails during cleanup:
http://www.osronline.com/showThread.cfm?link=67122
```
It is not advisable that you fail IRP_MJ_CLEANUP. If you want to protect files from being deleted do it in IRP_MJ_SET_INFORMATION/FileDisposition.
If you want to protect files from DELETE_ON_CLOSE, do it in IRP_MJ_CREATE. You should be able to set the DeletePending to FALSE in IRP_MJ_CLEANUP provided it was set TRUE in IRP_MJ_SET_INFORMATION.
But you won't be able to protect files which are opened by DELETE_ON_CLOSE because you have no way to tell the underlying file system to unset its internal flag.
Thus it is better to avoid doing your processing in IRP_MJ_CLEANUP dispatch and I strongly advise not failing it.
```
So here for dokan, if cleanup userland fail, we will return success anyway. The file will disappear from the explorer and will appear again when explorer refresh the list of files.

2. This need to be verified:
When a folder is deleted, the CreateFIle access of the folder used will denied other request to happen (sharing flag).
When a file is created, Windows make a CreateFile on the Folder to see if it exists, and it will normally fail since no sharing has been set previously by the Delete CreateFile.

I didn't test right now but thats how I remember.
The issue with the FUSE wrapper I can see here is that shared_mode is not really used looks like. BUT normally the kernel dokan check the shared access and will return directly an error during CreateFile on the Folder. So FUSE is protected of this behavior.

Rondom:
2. So, I did some test and there is indeed the possibility of races.
I tested the following two scenarios. I placed a `Sleep(15000);` [right before the return in `delete_directory`](url)
      * Fired rmdir with empty dir in one cmd-window, created file inside the directory from another cmd-window while rmdir was running => worked without a problem. rmdir returned success and the directory was not deleted.
       * Fired rmdir with non-empty dir in one cmd-window, removed the file inside the directory from another cmd-window while rmdir was running => worked without a problem, rmdir said that the directory was not empty

So, I understand that this is a separate issue that has to do with a flag given during `CreateFile`, right? So does this affect deleting files as well, then?
Given that it is a separate issue, that exists "anyway", do you think it makes sense to merge this PR?

(I do not have any idea on how to fix this without doing some reading, because, as you know, I do not have much of a background on Windows-FS-semantics.)

dokan_fuse/src/fusemain.cpp
```diff
@@ -424,15 +424,67 @@ int impl_fuse_context::create_directory(LPCWSTR file_name,
   return ops_.mkdir(fname.c_str(), dirmask_);
 }
 
+int impl_fuse_context::readdir_filler_set_has_files(void *buf, const char *name,
+                                      const struct FUSE_STAT *stbuf,
+                                      FUSE_OFF_T off) {
+  bool *has_files = static_cast<bool *>(buf);
+  int ret = 0;
+  if (strcmp(name, ".") && strcmp(name, "..")) {
+    //fprintf(stderr, "\tDirectory is not empty (found direntry %s).\n", name);
```
Rondom: What about this line? Remove it, uncomment it or leave as it is (commented)?
The whole file does not have any debug-prints, so the `FPRINTF`-macro defined in dokanfuse.cpp is not available.
If the dir is to be deleted it will still show `dispositionInfo->DeleteFile = 1` as the next printout in any case.

Liryna:
Hi @Rondom ,

Sorry for the delay, yes I think this PR has to anyway be merged.
During your test, did created the new file using the local path (C:\\) or the mirrored path dokan (M:\\) ? 

I have to think about it and see if even this could happen in the mirror C.
At least, we should handle the case that it is not possible to create a folder in a folder when it is open to delete.
Can you create another issue for this ? ❤️

dokan_fuse/src/fusemain.cpp
Liryna: @Rondom Yes, this can be removed I think.

Rondom:
Of course I am talking only about the mirror filesystem M:.

I think for testing it with the C-Mirror, you could add a `Sleep(15000);` [in the C-Mirror here](https://github.com/dokan-dev/dokany/blob/b91c54676413d7d972ff402f4241fd3f2edaeed3/samples/dokan_mirror/mirror.c#L769).

I am not familiar with Windows-FS-semantics to know whether this is by design. My gut feeling is that it should not be possible and some flag passed to ``CreateFile` would grant exclusive access to the process that wants to delete the directory, but I am not sure...

All in all this fixes PR one issue #270. The race-issue (if it is an issue) existed even before the fix. It becomes a bit more likely with the fix, but compared to the unfixed situation (always returns sucess, even if directory is not empty) it is an improvement.
What can happen after applying this PR is that
1. A race occurs, the delete returns success, but the directory is not deleted, because a new file is put into it (Exactly the same behaviour as before).
2. A race occurs, the delete fails (dir not empty). In the meanwhile the dir becomes empty. The dir is not deleted. This is probably a non-issue.
3. No race occurs, either sucess or dir-not-empty is returned and the directory is either deleted properly (sucess) or not (dir-not-empty). This is the desired behaviour.

So, after applying this PR, instead of showing the wrong behaviour in 100% of the cases, the wrong behaviour can still occur in X% of the cases, i.e. when a race occurrs.

Please tell me, if you agree/disagree.

Liryna:
Hi @Rondom ,

I agree ! With what you did the situation is fixed even if race condition could happen (very rare).
I will do the test with the mirror as you proposed with the sleep and continue the discussion on https://github.com/dokan-dev/dokany/issues/478 


Thanks a lot again for your contribution ❤️ !

Issue #465, opened 2017-02-27
Drive fails to mount on fuse-bindings example.js when contents are specified


sam0x17:
# Environment
 * Windows version: Windows 10 Pro (Virtualbox VM)
 * Processor architecture: x64
 * Dokany version: 1.0.2
 * Library type (Dokany/FUSE): FUSE via node.js `fuse-bindings` package (https://github.com/mafintosh/fuse-bindings)
 
# Check List
  - [X] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [X] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [X] I can always reproduce the issue with the provided description below.
  - [X] I have updated Dokany to the [latest version](https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
  - [x] I tested one of the last [snapshot](https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot) from appveyor CI

# Description

Using the default example from https://github.com/mafintosh/fuse-bindings/blob/master/example.js, with the `gid` and `uid` parameters excluded (since `process.getgid()` and `process.getuid()` are unsupported on windows and I have been unable to find a way of specifying the windows SID equivalents via Node), dokany will fail to mount. It works if I leave the drive empty, but the second I uncomment the line creating the 'test' contents (`if(path === '/') return cb(0, ['test'])`), it will fail, though no BSOD occurs. At least one other user is experiencing this identical issue. Issue on fuse-bindings: https://github.com/mafintosh/fuse-bindings/issues/40.

# Logs

Full kernel debug output: http://pastebin.com/raw/A7p7EqhZ

I will gladly provide additional output, just tell me what to do!

samkelly:
Possibly of note excerpts from kernel log (in chronological order):

```
00055627	0.00486348	[DokanFS]    => DokanDiskDeviceControl	
00055628	0.00486576	[DokanFS]    DiskDeviceControl Device name \Device\Volume{d6cc17c5-1731-4085-bce7-964f1e9f5de9} 	
00055629	0.00486690	[DokanFS]   IOCTL_STORAGE_QUERY_PROPERTY	
00055630	0.00486842	[DokanFS]     PropertyStandardQuery Unknown 6	
00055631	0.00486918	[DokanFS]    <= DokanDiskDeviceControl	
[...]

PR #475, opened 2017-03-12
Ensure fuse context is set during walk directory

marinkobabic:
When fuse bindings calls the filler which is the dokans walk_directory method. The walk_directory is called on different thread. We need to ensure that even then the fuse context is set. This fixes one part of the issue https://github.com/dokan-dev/dokany/issues/465. The other changes are inside of fuse-bindings required.


Liryna:
Thank you @marinkobabic for the changes !!!! ❤️

PR #481, opened 2017-03-21
may cause memory leaks

baixiangcpp:


Liryna:
Hi @baixiangcpp ,

This is not a leak. The buffer is cleaned in ReleaseDokanOpenInfo https://github.com/dokan-dev/dokany/blob/1d8aa517ef86b3fbef6c6c8ac03257f9c4e5de2d/dokan/dokan.c#L598
BUT during time that I have review your PR, if found that indeed the DirListHead is not cleaned all case since the last PR that changed DispatchDirectoryInformation

See my change: https://github.com/dokan-dev/dokany/commit/a3910ccb03cb05355ce2b746492f4a9b054c9685

Thanks for your PR 👍

PR #482, opened 2017-03-21
flush the targetObject as first and then the source file

marinkobabic:
Before rename the targetObject should be flushed as well


Liryna:
Thanks a lot @marinkobabic 👍

PR #484, opened 2017-03-25
add DokanNotifyReportChange when  DokanCompleteWrite

baixiangcpp:
ReadDirectoryChanges cannot work without this

### DIFF
--- a/sys/cleanup.c
+++ b/sys/cleanup.c
@@ -173,6 +173,12 @@ VOID DokanCompleteCleanup(__in PIRP_ENTRY IrpEntry,
   status = EventInfo->Status;
 
   DokanFCBLockRO(fcb);
+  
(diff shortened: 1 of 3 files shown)


baixiangcpp:
when notepad++ write something to a file, ReadDirectoryChanges  cannot  monit  wirte event.

Liryna:
Hi @baixiangcpp ,

Thank you for the PR !
I understand the need to add `FILE_NOTIFY_CHANGE_LAST_WRITE` that is missing.
But it should not be in write as you proposer but more in cleanup since the notification is here to say that last write happened. Probably we need to add a flag to ccb during write and check it during cleanup.

Liryna:
Hi @baixiangcpp 

I have include your changes in https://github.com/dokan-dev/dokany/commit/db9453b1777d0f1fde6501f60476ee77e5738b39
but I failed the commit author...it should be you sorry 😢 

Thanks again for your contribution !

PR #485, opened 2017-03-29
CI: Set AppVeyor build version

Rondom:
This sets the AppVeyor build version according to the version we get from the repository, i.e. `1.0.2-590` becomes ``1.0.2.1590` (like the installer and DLL-versions.


Rondom:
This produces broken links like 
https://ci.appveyor.com/project/Maxhy/dokany/build/1.0.2-163 while the build is running. Only when the build is finished the link is updated on the GitHub side to for example 
https://ci.appveyor.com/project/Maxhy/dokany/build/1.0.2.1163

So, don't merge this. :-(

Alternative approach
1. Keep `version` in appveyor.yml in sync with "our version" (i.e. format). In case of mismatch update (GitHub links will be broken, during build)
2. Keep `version` in appveyor.yml in sync with "our version" (i.e. format). In case of mismatch warn
3. Keep `version` in appveyor.yml in sync with "our version" (i.e. format). In case of mismatch fail build
4. Leave everything as it is
So don't merge this as it is.

Liryna:
Hi @Rondom ,

Github broken link during the build is alright I think. I personally always wait at the end if everything passed or otherwise look in the history current running build.
But maybe we can ask appveyor why `Update-AppveyorBuild` does not directly update in live the link ?

Rondom:
> But maybe we can ask appveyor why `Update-AppveyorBuild` does not directly update in live the link ?

appveyor/ci#1074

timofonic:
@Rondom @Liryna 
Any news about this? Any reason this is still open? Can someone reply about this? Thanks in advance :)

Rondom:
@timofonic Not sure what your intention is in inquiring about this issue, given that it does not really affect users in any way. :-)

The long story is that I started developing a cleaner approach for the version-handling, but did not finish it. https://github.com/Rondom/dokany/tree/no_hardcode_version

Do you want to help?

PR #489, opened 2017-04-03
MirrorCreateFile - Return the correct error code when trying to open a directory as file.

### LINKED ISSUE #488: Mirror opening File as Directory on Windows Share when Office Word 2010 Opening the file
# Environment
 * Windows version (Client):  Windows 7
 * Windows version (Server):  Windows 10
 * Processor architecture: amd64
 * Dokany version: 1.0.2.1136 Snapshot
 
# Check List
  - [] I have updated Dokany to the [latest version](https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description
The Mirror would trying to open a file as directory when the Office Word 2010 Opening the file on Windows Share. This seems only happen on mounted the remote drive and the local drive doesn't seems get any error.

# Logs

> ###Create 0189
> 	eventInfo.SerialNumber = 4A3, eventInfo.Status = 0, origFileName = (null) CreateFile : \\192.168.200.18\share\
>   AccountName: VMBB-PC4, DomainName: VMBB-PC20-PC
> 	ShareMode = 0x1
> 	FILE_SHARE_READ
> 	DesiredAccess = 0x120080
> 	FILE_READ_ATTRIBUTES
> 	READ_CONTROL
> 	SYNCHRONIZE
> 	STANDARD_RIGHTS_READ
> 	STANDARD_RIGHTS_WRITE
> 	STANDARD_RIGHTS_EXECUTE
> 	FlagsAndAttributes = 0x0
> 	OPEN_EXISTING
> 	error code = 123
> 
> 
> CreateFile status = c0000033

This image show the mirror is trying to open as a file when GetFileAttributes returned 0x00000010 (FILE_ATTRIBUTE_DIRECTORY) .

open_directory_as_file_issue

yin19941005:
Returning the correct error code when trying to open a directory as file. The STATUS_FILE_IS_A_DIRECTORY NTSTATUS will be returned.

This should fix #488 .

### DIFF
--- a/samples/dokan_mirror/mirror.c
+++ b/samples/dokan_mirror/mirror.c
@@ -366,6 +366,14 @@ MirrorCreateFile(LPCWSTR FileName, PDOKAN_IO_SECURITY_CONTEXT SecurityContext,
         CreateDisposition == FILE_CREATE)
       return STATUS_OBJECT_NAME_COLLISION; // File already exist because
                                            // GetFileAttributes found it
+
+	// Check first if we're trying to open a directory as a file.
+	if (fileAttr != INVALID_FILE_ATTRIBUTES &&
+		(fileAttr & FILE_ATTRIBUTE_DIRECTORY) &&
+		!(CreateOptions & FILE_DIRECTORY_FILE)) {
+		return STATUS_FILE_IS_A_DIRECTORY;
+	}
+
     handle = CreateFile(
         filePath,
         genericDesiredAccess, // GENERIC_READ|GENERIC_WRITE|GENERIC_EXECUTE,


Liryna:
Hi @yin19941005 ,

Thank you for the fix ! This seems to be the correct way to do it :+1: 
Does CreateOptions has the flag `FILE_NON_DIRECTORY_FILE` ?

yin19941005:
Hi @Liryna ,

Yes, the CreateOption has the flag `FILE_NON_DIRECTORY_FILE`. I remember that there was a fix to preventing mirror opening a directory as a file (#370) . So I think this is similar with that bug and apply similar fix to it.

`FILE_NON_DIRECTORY_FILE` is 0x00000040 when the CreateOption get 0x60 flag : 
open_directory_as_file_issue2

Liryna:
Thank you @yin19941005 for the debug !
So what do you think about using directly `FILE_NON_DIRECTORY_FILE` instead of `!(CreateOptions & FILE_DIRECTORY_FILE)` ?
I am afraid with your current test that in some case even if `FILE_DIRECTORY_FILE` is not set, kernel would probably expect that it works for directory (not in mind a case but better to use `FILE_NON_DIRECTORY_FILE` directly).

yin19941005:
Hi @Liryna ,

It's a good idea to use `FILE_NON_DIRECTORY_FILE` directly instead of `!(CreateOptions & FILE_DIRECTORY_FILE)`! We could not come up with any cases in mind but sounds more reliable if we are going to use `FILE_NON_DIRECTORY_FILE` this flag.

I will make a new pull request by using `CreateOptions & FILE_NON_DIRECTORY_FILE` this condition.

Issue #490, opened 2017-04-04
FILE_NAMED_STREAMS flag


bailey27:
I noticed that MirrorGetVolumeInformation() never sets the FILE_NAMED_STREAMS flag, even when the underlying filesystem (e.g. NTFS) supports them.

It looks like Alternate Data Streams generally work.  Mirror passes all the streams tests in winfstest.

However, without this flag, if you copy or move a file with an alternate data stream on a mirror drive,  then you lose the alternate data stream, because the OS doesn't know that the fs supports streams.

Also, if you do set this flag, then you start getting weird paths with an extra backslash after the file name and before the stream specifier.

I mean like 

```
\foo\boo\foo.txt\:blah:$DATA
```

And these paths don't work unless you get rid of the backslash that's in front of the first colon.

I think GetFilePath() needs to take care of that.

I found this web site that has some tools for inspecting streams in files.

http://www.flexhex.com/docs/articles/alternate-streams.phtml

Liryna:
Hi @bailey27 ,

Oh yes `FILE_NAMED_STREAMS` seems to be missing !
I guess the streams are currently working because windows kernel do not refuse stream request even if flag is missing but let the driver handle or not CreateFile for streams (like we are doing: https://github.com/dokan-dev/dokany/blob/d3b7927fbc68296f9c41228a3e5fabe322be0b2a/sys/create.c#L655-L661).
BUT having `FILE_NAMED_STREAMS` seems to be checked by some applications (like explorer here) to enable features (copy extra streams).

Do you know if the path `\foo\boo\foo.txt\:blah:$DATA` come from the kernel or it is dokan kernel that wrongly changed it ?

bailey27:
I think maybe it's happening in the driver in create.c in DokanDispatchCreate() because of needBackSlashAfterRelatedFile.  I'm not sure why it thinks it needs to add a backslash in this case.

I only looked at the code.  I haven't confirmed this.

bailey27:
It looks like that is it.

I think when the file is a normal file, not a stream, then the relatedFileName is the full path of the directory that the file is in, and  fileObject->FileName is the name of the file itself.

But when the file is a stream, then the relatedFileName is the full path to the file, and fileObject->FileName is the stream name.

If I change it like this, then copying and moving a file with an alternate data stream keeps the stream intact.

```
diff --git a/sys/create.c b/sys/create.c
index 4c4f7ed..42fa37d 100755
--- a/sys/create.c
[...]

PR #493, opened 2017-04-06
Streams flag

bailey27:
For issue #490: FILE_NAMED_STREAMS flag 

set FILE_NAMED_STREAMS in FileSystemFlags  MirrorGetVolumeInformation() in mirror.c.

Do not append slash to path in front of filename if filename is a stream (DokanDispatchCreate() in create.c).
Also, do not prepend path (\) in front of filename if file is really an alternate data stream of the root directory.


Liryna:
Thanks a lot @bailey27 again for your work ❤️ always appreciate !

Issue #476, opened 2017-03-15
Could not save MS Word files normally on mirror.exe mounted Windows Share Drive (SMB share drive)


yin19941005:
# Environment
 * Windows version (client): Windows 7 (64 bit)
 * Windows version (smb server): Windows 10 (64 bit)
 * Processor architecture: amd64
 * Dokany version: 1.0.2.1000
 * Library type (Dokany/FUSE): Dokany
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the [latest version](https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
  - [ ] I tested one of the last [snapshot](https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot) from appveyor CI

p.s. I didn't download the snapshot from appveyor CI but I compiled [this commit](https://github.com/yin19941005/dokany/commit/e43caaf645b5772174c0c314047f8f7fbcc1f8b5) by adding more DbgPrint to the latest commit as debug build and the problem persist. 

# Description
I had used the command below to mount up the drive by mirror : 
Mount Local Drive Command : `mirror.exe /r C:\dokan_temp /l Z /d /s /t 1`
Mount SMB Share Drive Command : `mirror.exe /r \\192.168.200.18\share /l Z /d /s /t 1`

The local drive doesn't come with any problem. I provide the command just because I will provide the normal log for reference below. I am able to save the word document normally.

The OpLocks issue occurs on SMB Share Drive. I mounted the drive with command above then I copied an existing `.docx` file and start edit it by Microsoft Word 2010 (MS Word 2010). When I going to save it (Press Save button or Ctrl+S), I can't save the file and it prompt up "Save As.." Dialog. After the "Save As.." dialog prompt up, the behavior will depend on the dokany version you are using.

For dokan-1.0.2.1000 : 
After the "Save As.." Dialog prompt up, I can't save the file at all. When I tried to save to the exact same file, the warning message for permission error prompt up as below and failed to save the file. I tried to save to another name, let's say "Testing002.docx", but I am still failed to save the fail.

[...]

PR #507, opened 2017-05-11

yin19941005:

This fix would fix part of the issue on #476 and save the file correctly.

### DIFF
--- a/samples/dokan_mirror/mirror.c
+++ b/samples/dokan_mirror/mirror.c
@@ -1095,6 +1095,7 @@ static NTSTATUS DOKAN_CALLBACK MirrorGetFileSecurity(
     PULONG LengthNeeded, PDOKAN_FILE_INFO DokanFileInfo) {
   WCHAR filePath[DOKAN_MAX_PATH];
   BOOLEAN requestingSaclInfo;
+  BOOL isGetUserObjectSecuritySuccessful = FALSE;
 
   UNREFERENCED_PARAMETER(DokanFileInfo);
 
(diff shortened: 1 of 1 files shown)


sefai:
Hello @yin19941005, 

Thank you for this great effort. I don't want to look smart but there may be a more clean implementation?

```
if (!GetUserObjectSecurity(handle, SecurityInformation, SecurityDescriptor,  BufferLength, LengthNeeded)) {
   int error = GetLastError();
   if (error == ERROR_INSUFFICIENT_BUFFER) {
      DbgPrint(L"  GetUserObjectSecurity error: ERROR_INSUFFICIENT_BUFFER\n");
      CloseHandle(handle);
      return STATUS_BUFFER_OVERFLOW;
   } else {
      DbgPrint(L"  GetUserObjectSecurity error: %d\n", error);
      CloseHandle(handle);
      return DokanNtStatusFromWin32(error);
   }
 }
 else {
   DWORD securityDescriptorLength = GetSecurityDescriptorLength(SecurityDescriptor);
   DbgPrint(L"  isGetUserObjectSecuritySuccessful is true,  *LengthNeeded = securityDescriptorLength \n");
   *LengthNeeded = securityDescriptorLength;
}

CloseHandle(handle);
```

What do you think?

yin19941005:
Hi @sefai ,

Thank you for your opinion! Yes, I think your implementation looks more clean! I didn't study the if-else for `ERROR_INSUFFICIENT_BUFFER` carefully before and I thought we need a Boolean but actually not. 

I would recreate another pull request for this implementation.

PR #508, opened 2017-05-11
Fix typo in recent commit

yin19941005:
Fix typo in recent commit.


Liryna:
Hi @yin19941005 ,

Thank for the fix !

PR #509, opened 2017-05-12

yin19941005:

This fix would fix part of the issue on #476 and save the file correctly. Re-implement #507  to look more clean.

### DIFF
--- a/samples/dokan_mirror/mirror.c
+++ b/samples/dokan_mirror/mirror.c
@@ -1156,6 +1156,12 @@ static NTSTATUS DOKAN_CALLBACK MirrorGetFileSecurity(
       return DokanNtStatusFromWin32(error);
     }
   }
+  
+  // Ensure the Security Descriptor Length is set
+  DWORD securityDescriptorLength = GetSecurityDescriptorLength(SecurityDescriptor);
+  DbgPrint(L"  isGetUserObjectSecuritySuccessful is true,  *LengthNeeded = securityDescriptorLength \n");
+  *LengthNeeded = securityDescriptorLength;
+
   CloseHandle(handle);
 
   return STATUS_SUCCESS;


Liryna:
Hi @yin19941005 ,

Thank you for the PR.
Can you just change the debug msg that still print the boolean variable ?

yin19941005:
Hi @Liryna ,

Do you mean I should change the debug message `isGetUserObjectSecuritySuccessful is true` to `GetUserObjectSecurity return true` or something similar with that? Because in this implementation I haven't add Boolean variable. Below is [the code](https://github.com/dokan-dev/dokany/blob/master/samples/dokan_mirror/mirror.c#L1146) :

```
if (!GetUserObjectSecurity(handle, SecurityInformation, SecurityDescriptor,
                              BufferLength, LengthNeeded)) {
     int error = GetLastError();
     if (error == ERROR_INSUFFICIENT_BUFFER) {
       DbgPrint(L"  GetUserObjectSecurity error: ERROR_INSUFFICIENT_BUFFER\n");
       CloseHandle(handle);
       return STATUS_BUFFER_OVERFLOW;
     } else {
       DbgPrint(L"  GetUserObjectSecurity error: %d\n", error);
       CloseHandle(handle);
        return DokanNtStatusFromWin32(error);
      }
    }
```

As @sefai remind, if GetUserObjectSecurity return false and error == ERROR_INSUFFICIENT_BUFFER, will return STATUS_BUFFER_OVERFLOW. If GetUserObjectSecurity return false and error != ERROR_INSUFFICIENT_BUFFER, will return DokanNtStatusFromWin32(error). 

That means all the cases with GetUserObjectSecurity return false will be returned before the calling the GetSecurityDescriptorLength function. In other words, it is not meaningful to add a Boolean variable to it. 

I agree with @sefai so I removed the Boolean variable.

Liryna:
@yin19941005 I totally agree with this 👍 I was meaning the debug message was still `isGetUserObjectSecuritySuccessful is true`

yin19941005:
OK, I would change the debug message and open up another pull request. Sorry for didn't change the debug message correctly.

PR #510, opened 2017-05-12

yin19941005:

This fix would fix part of the issue on #476 and save the file correctly. Re-implement #507, #509 to look more clean.


Liryna:
Thank @yin19941005 for the quick changes!

Issue #517, opened 2017-05-21
Mirror driver /c 'current session' option doesn't appear to work in Terminal Services Session


shashiu:
Feature request can skip this form. Bug report **must** complete it. `Check List` must be 100% match or it will be automatically closed without further discussion. Please remove this line.

# Environment
 * Windows version: Windows 2012
 * Processor architecture: x64
 * Dokany version: 1.0.3
 * Library type (Dokany/FUSE): Dokany 
 
# Check List
  - X[ ] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [X ] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [X ] I can always reproduce the issue with the provided description below.
  - [X ] I have updated Dokany to the [latest version](https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
  - [ ] I tested one of the last [snapshot](https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot) from appveyor CI

# Description
If I try to mount a folder using the mirror sample using the following command line on 2 separate terminal server sessions (logging in on the same server using the RDP client mstsc.exe) I get a 'drive letter already used' error. Mirror command line : "mirror.exe /n /c /r c:\Users /l m". The command line works fine in the first session, but gives me an error saying "Drive something wrong" in the 2nd session. This is easily reproducible on any Windows Server 2012 R2 with Remote Desktop Enabled.

# Logs
Please attach in separate files: mirror output, library logs and kernel [logs](https://github.com/dokan-dev/dokany/wiki/How-to-Debug-Dokan).
In case of BSOD, please attach minidump or dump analyze output.

shashiu:
I am including a patch for the fix for the problem above. Basically, the FindMountEntry() routine is returning the wrong entry when 2 terminal services (remote desktop) sessions use the same drive letter. The fix is to include a session identifier in the DOKAN_CONTROL structure and use that in the comparison in FindMountEntry.
[RDS-fix.patch.txt](https://github.com/dokan-dev/dokany/files/1033940/RDS-fix.patch.txt)

shashiu:
@marinkobabic , please find the fix for the issue above.

Liryna:
Hi @shashiu ,

Thank you for the patch and for taking the time to look into this issue !
I think this is the correct way to fix the issue and this seems to not break kernel API.
[...]

Issue #332, opened 2016-09-09
Drive letter cannot be reclaimed after killing process with current_session option specified


kenshinsh:
# Environment
- Windows version: windows 7 64-bit
- Processor architecture: x64
- Dokany version: 1.0.0 rc4
- Library type (Dokany/FUSE): Dokany
# Check List
- [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
- [x] I can always reproduce the issue with the provided description below.
- [x] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
# Description

Hello,
I noticed an issue that the network share does not get released by the operating system when the user space process gets killed off via the task manager when user_current option is specified.

To reproduce the issue:
1. run command: mirror.exe /r D:\test -l p /d  /s /n /c /u \test\share
2. observed the network share is mounted
3. open task manager and kills mirror.exe
4. re-run the same mirror.exe command and observes that the same drive letter is no longer available

I have tried the same step without /u and I was able to reclaimed the same drive letter after a few seconds of wait time.

I have attached the mirror.exe output with and without current_session specified.
# Logs

[mirror_output_with_current_session.txt](https://github.com/dokan-dev/dokany/files/464987/mirror_output_with_current_session.txt)
[mirror_output_without_current_session.txt](https://github.com/dokan-dev/dokany/files/464990/mirror_output_without_current_session.txt)

Liryna:
Hi @kenshinsh ,

Thank you for your report !

I have been able to reproduce your issue.
You don't specialy need the network option to make it happen. Mounting a device with current session and killing it does not clean the mount point as it does with global mount point.

Maxhy:
Looks like this is coming from https://github.com/dokan-dev/dokany/blob/2c6ca3a6594741d0edc5c21f7116ffffee9fa579/sys/timeout.c#L326 as the timeout thread is running in system context.
Not sure about how we should handle this specific case. Maybe by storing the global dosdevice path with session information (eg. `\Sessions\0\DosDevices\M`) on device extension, but I don't think kernel mode naming supports it, neither how to get it if it does. @marinkobabic?

marinkobabic:
Option 1:
[...]

Issue #312, opened 2016-08-06
Dokanctl - Allow unmount option for current user volumes


Liryna:
`/u` option require admin rights for now.
It would be nice to be able to unmount current user volumes `DOKAN_OPTION_CURRENT_SESSION`.

There is currently no possibility to check volume owners and therefore see if the user has the right to unmount them.
The easiest solution would be to retrieve SymbolicLink owner information (if that’s possible) and see if it matches the current user.

marinkobabic:
If Dokanctl does not need admin rights anymore (think, this has been fixed) then the unmount of current session drives should be possible using the following pull request https://github.com/dokan-dev/dokany/pull/523 I did not test it, but it must work. It will grab the current session from the request and so unmount the correct device.

infeo:
Hey, is there any progress in this issue? because i would really like to use dokany, but i don't have always admin rights. Currently in windows 10 and the recent dokany binaries (from 27.11.2017) it still needs admin rights to unmount my drive.

Liryna:
Hi @infeo ,

Using dokanctl to unmount is only needed if your application did not implement a unmount mechanism by calling dokan API for it (see documentation).
Dokanctl is really only needed if you do not know who has mounted the device (software/user) and you force unmount.

the-Arioch:
Hello, can you give a shortcut, if dokanctl (or Windows Explorer) tries to unmount a disk, what would it look application-wise, will there be some event oushed from dokan driver to the applicatoin, some flag set to be polled?

https://github.com/DDoSolitary/yasfw/issues/28

PR #523, opened 2017-05-30
Fix current session issues

marinkobabic:
The following issues should be fixed

https://github.com/dokan-dev/dokany/issues/517
https://github.com/dokan-dev/dokany/issues/332

and the following one may also work properly for session drives
https://github.com/dokan-dev/dokany/issues/312

### DIFF
--- a/dokan/dokan.c
+++ b/dokan/dokan.c
@@ -112,6 +112,8 @@ BOOL CheckDriveLetterAvailability(WCHAR DriveLetter) {
   dosDevice[4] = driveLetter;
   driveName[0] = driveLetter;
 
+  DokanMountPointsCleanUp();
+
   if (!IsValidDriveLetter(driveLetter)) {
     DbgPrintW(L"CheckDriveLetterAvailability failed, bad drive letter %c\n",
               DriveLetter);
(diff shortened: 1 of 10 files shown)


Rondom:
Did you intentionally close this one?

marinkobabic:
Yes. I have created a new one.

Am 30.05.2017 23:10 schrieb "Andreas Gnau" <notifications@github.com>:

Liryna:
Hi @marinkobabic ,

Sorry for the delay, I took a look at the PR, this seems to be good !
I will just take a little more time to test to the issue related to it and see if it can also be closed.

Thank again for your work ! ❤️

Liryna:
Hi @marinkobabic ,

I tried to reproduce all the issues:
https://github.com/dokan-dev/dokany/issues/517 : I could not test this case exactly but it should work 👍 
https://github.com/dokan-dev/dokany/issues/332 : Yes it works! the drive letter can be reused ! The only strange behavior is that the device is still in the explorer and can be access but empty after killing the mirror.
https://github.com/dokan-dev/dokany/issues/312 : User A can dismount user B drive. Maybe here we should detect in the kernel if the request comes from an admin or simple user and give only rights to admin to dismount all devices.

marinkobabic:
User A should not be able to unmount the drive of the user B if both use
the current session option. The will have different windows sessions.

Am 05.06.2017 13:42 schrieb "Liryna" <notifications@github.com>:

shashiu:
Is there going to be a new build with this fix ?

Liryna:
@shashiu There is every time a snapshot build made by the CI selfsigned
https://ci.appveyor.com/api/buildjobs/ss79dhy6ah90bfqj/artifacts/DokanSetup-1.0.3.1193.exe

https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot

It would be nice if you could see if this fix your issue before making a fix release official (that take some times).

shashiu:
@Liryna<mailto:notifications@github.com>, I checked out the build and it seems to fix the issue with remote desktop sessions, that I was seeing before. I'll keep testing it and let you know if I find any new issues.


Thanks.


________________________________

Liryna:
Hi @marinkobabic ,

Sorry for the long delay 😢 I merged this only now.
Thanks a lot again for your work !

Issue #519, opened 2017-05-26
BSOD on IFSTest on AppVeyor Win2012R2 image


Rondom:
# Environment
 * Windows version: 2012 R2
 * Processor architecture: amd64
 * Dokany version: master (0e159e37d28d44828210d72b34d736ccd43933d5) with some AppVeyor changes
 * Library type (Dokany/FUSE): dokany
 
# Check List
  - [ x ] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [ x ] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [ x ] I can always reproduce the issue with the provided description below.
  - [ x ] I have updated Dokany to the [latest version](https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
  - [ x ] I tested one of the last [snapshot](https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot) from appveyor CI

# Description
Running IFSTest (Win10Anniversary) on the AppVeyor VS2015 image causes a BSOD.
I have not tried to reproduce it anywhere else. The corresponding commit, has a modified mirror_test.ps1 that runs IFSTest.
(I will also bug the AppVeyor-people with two reproducible test cases for BSOD, in one situation the current build step will silently be skipped, in the other one the build hangs completely...)

# Logs
Can be found in the [AppVeyor output](https://ci.appveyor.com/project/Rondom/dokany/build/1.0.3-339/job/nj3gfa14hf35r4t8#L1197). A memory.dmp can be downloaded there as well.

(I am also wondering why KD is not displaying any line-numbers in the AppVeyor-output. Maybe someone can tell me how to load the debug-symbols currectly.)

    & "C:\Program Files (x86)\Windows Kits\10\Debuggers\x64\kd.exe" -z $memory_dmp -y "SRV*${env:APPVEYOR_BUILD_FOLDER}\x64\Win8.1Release\*http://msdl.microsoft.com/download/symbols" -c "!analyze -v; q" }

Liryna:
Hi @Rondom ,

Thank you trying to add IFSTest to appveyor 👍 

Should probably change the build configuration to build as debug and not release 😢 
@marinkobabic any idea what release pdb is not used ?

marinkobabic:
Normally winding displays why the pub is not used. It shows if the pub has
been found and the reason of not using it.

[...]

PR #529, opened 2017-06-12
Fix for CreateFileMapping 

sefai:
Save file size to later know if MmCanFileBeTruncated call is needed or not.

### DIFF
--- a/sys/fileinfo.c
+++ b/sys/fileinfo.c
@@ -299,13 +299,6 @@ VOID DokanCompleteQueryInformation(__in PIRP_ENTRY IrpEntry,
     status = EventInfo->Status;
 
     if (NT_SUCCESS(status)) {
-      if (irpSp->Parameters.QueryFile.FileInformationClass ==
-          FileAllInformation) {
-        PFILE_ALL_INFORMATION allInfo = (PFILE_ALL_INFORMATION)buffer;
-        allInfo->PositionInformation.CurrentByteOffset =
-            IrpEntry->FileObject->CurrentByteOffset;
-      }
-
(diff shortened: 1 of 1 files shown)


Rondom:
Can we run runkarr on this PR to see if it improves things compared to the current implementation?
See #519 for a link to a branch that has an installation script for the MS filesystem testing utilities.

Rondom:
https://github.com/Rondom/dokany/tree/ifstest/scripts

sefai:
Hello @Rondom 
Do you want me to install MS filesystem testing utilities locally and execute the tests? Isn't it possible to run that IFSTest on Appveyor?

Rondom:
I will do it at a later point. I only thought that a this can be helpful that everything is all right, but I just noticed that it requires additional work. I will add it to AppVeyor at a later point.

Liryna:
Hi @sefai ,

Thanks a lot for the PR !
All seems to be correct but could you just add some fcb lock when accessing the FileSize ?
This should not create deadlock but worst to retry with it 👍

sefai:
Hello @Liryna 

I have tried locking FCB when accessing the FileSize, but unfortunately it creates a deadlock.

Also, my PR is not correct, the if clause should be negated 

`if (!(irp->Flags & IRP_PAGING_IO)) `

Can you do that, or should I create a new PR?

Liryna:
@sefai I have add a lock to the filesize set value and rebase your commit
https://github.com/dokan-dev/dokany/commit/38348f5fe160e9f2d384b992c69eed8b8d8e1824

Thank again for your time and sorry for the delay !

PR #531, opened 2017-06-13
Prevent BSOD if output buffer address is not in user space

marinkobabic:
See https://www.greyhathacker.net/?p=818


Liryna:
Hi @marinkobabic ,

Thanks a lot for the quick fix and well found ! 👍

PR #534, opened 2017-06-18
Run IFSTest on AppVeyor

Rondom:
This was more work than expected, but here it is:
* add install-script install_wlk_ifstest.ps1 for installing the IFSTest part of WLK/HCK. Currently only Win8.1 and Win10 Anniversary versions are supported. Not really sure about backward/forward compatibility of installing them on other OSes, but I think nothing bad should happen except failing tests (see the source on how to support other version or change the conditions for choosing one version over the other)
* run IFSTest from mirror_test.ps1
* Make AppVeyor re-use the installer built in the All-"Configuration", run the tests using that installer on both Win2012R2 (8.1) and and Win2016 (10 Anniversary)
* Please note that the automatic BSOD analysis does not work at the moment on AppVeyor's Win2016 (VS-2017) image, because they forgot to install windbg. You are free to download the installer and the memory.dmp from the artifacts and run windbg locally. appveyor/ci#1615

Minimum checklist before merging:
[ x ] No BSODs such as #519
[ x ] Issues filed for every skipped test that is viable to fix (see mirror_test.ps1) and issue mentioned beside the skipped test in the source code

Nice to have (maybe in a separate PR):
* AppVeyor workarounds not necessary
* Putting results into the tests-section of AppVeyor
* Calculate delta between commits e.g. when submitting a PR, compare passed/failed/skipped tests to master and post a comment in the GitHub issue.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,6 +11,7 @@
       CYG_CACHE: '%DOKAN_CI_CACHE%\cygwin'
       MSYS2_CACHE: '%DOKAN_CI_CACHE%\msys2'
       CHOCO_CACHE: '%DOKAN_CI_CACHE%\choco'
+      WLK_INST_CACHE: '%DOKAN_CI_CACHE%\wlk_inst'
 
   os: Visual Studio 2015
   version: 1.0.3-{build}
@@ -18,6 +19,7 @@
   - All
   - Coverity
   - FsTest
+
   platform:
   - x64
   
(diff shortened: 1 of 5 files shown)


Liryna:
This is beautiful @Rondom !!! 🏆 🍾 
Now lets hunt the crash !

Rondom:
Once the build passes, we can merge this, I think.

Liryna:
All seems to be perfect for me @Rondom ! 🏆 extreme great job here 😃 !

PR #541, opened 2017-06-25
Prevent BSOD if output buffer address is not in user space

marinkobabic:
Fix BSOD according to https://www.greyhathacker.net/?p=818


marinkobabic:
Have no my discs there to be able to do the tests myself. Let's see if the system reboots now because of BSOD. Warnings should be gone.

marinkobabic:
it seems like i have to test it at home

PR #553, opened 2017-07-12
Fix BSOD during searching the backslash

### LINKED ISSUE #549: Random BSOD when saving MS Word files on Windows shared Dokan drive
# Environment
 * Server : Windows version: Windows Storage Server 2016
 * Client : Windows 7
 * Processor architecture: AMD64 (for both client and server)
 * Dokany version: 1.0.3.1000 release
 * Microsoft Office version: MS Word 2007
 
# Check List
  - [ ] My issue is valid with mirror default sample and not specific to my user-mode driver implementation (I didn't test with mirror because I think the issue occur in dokan1.sys)
  - [x] I can always reproduce the issue with the provided description below. But this takes some time.
  - [ ] I tested one of the last [snapshot](https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot) from appveyor CI

# Steps to reproduce : 

1. On the client VM, accessing the dokan drive, which mounted and shared out on the Server, create or open a MS Word file.

2. Modify the MS Word file and save it.

3. Repeat Step 2, until BSOD. Sometimes, the dokan will hold up. But wait for a timeout (maybe 3 minutes?), it can function again. Usually, I can reproduce the BSOD within 10 minutes.

# Description
The BSOD message is PAGE_FAULT_IN_NONPAGED_AREA (50). The causes of BSOD is on [this line of code](https://github.com/dokan-dev/dokany/blob/master/sys/dokan.c#L478). 

```
USHORT nameOffset;

... ... (Skip)

// search the last "\"
nameOffset = (USHORT)(FileName->Length / sizeof(WCHAR) - 1);
for (; FileName->Buffer[nameOffset] != L'\\'; --nameOffset)
;
```

As the `Locals` image on below you can see the FileName (PUNICODE_STRING) didn't contain any `\` character. The BSOD occur when `nameOffset = 0xFFFF` and trying to reference `FileName->Buffer[nameOffset]`. The reason why `nameOffset = 0xFFFF` is because the characteristics of the `Unsigned type`. I found a reference in [stackoverflow](https://stackoverflow.com/questions/7221409/is-unsigned-integer-subtraction-defined-behavior) : 

> A computation involving unsigned operands can never overflow, because a result that cannot be represented by the resulting unsigned integer type is reduced modulo the number that is one greater than the largest value that can be represented by the resulting type. (ISO/IEC 9899:1999 (E) §6.2.5/9)

So, when we performing `0 - 1` in `USHORT`, it will equals to `0xFFFF`. And this is dangerous and caused the BSOD. If the BSOD doesn't occur, it should be in the infinity loop and that's why it will hold up for some cases.

# Suggestion Fix

In my opinion, I think the normal case the `FileName` will contain the absolute path and we can find the `\` character but the MS Word 2007 some how some way generate an abnormal case. So, I suggest to change the code like this way : 

```
// 0 > 0 will return false and end the loop
while(nameOffset > 0)
{
	nameOffset--;

	if(FileName->Buffer[nameOffset] == L'\\') 
	{
		nameOffset++; // the next is the begining of filename
		break;
	}
}

nameOffset *= sizeof(WCHAR); // Offset is in bytes
```

I searched `] != L'\\'; --` on the entire solution and found these similar implementation may also cause the BSOD : 

[https://github.com/dokan-dev/dokany/blob/master/sys/dokan.c#L478](https://github.com/dokan-dev/dokany/blob/master/sys/dokan.c#L478)
[https://github.com/dokan-dev/dokany/blob/master/sys/event.c#L625](https://github.com/dokan-dev/dokany/blob/master/sys/event.c#L625)
[https://github.com/dokan-dev/dokany/blob/master/sys/timeout.c#L56](https://github.com/dokan-dev/dokany/blob/master/sys/timeout.c#L56)

@Liryna , do you think the suggestion fix is appropriate? Besides, are there other similar implementation may cause BSOD I couldn't found ?

# Logs
[analyze -v.txt](https://github.com/dokan-dev/dokany/files/1129890/analyze.-v.txt)

FAULTING_IP: 
dokan1!DokanNotifyReportChange0+123 [e:\islog\dev\app\dokan\dokany\sys\dokan.c @ 478]
fffff800`5c9c8ea3 0fb70441        movzx   eax,word ptr [rcx+rax*2]

PAGE_FAULT_IN_NONPAGED_AREA (50)
Invalid system memory was referenced.  This cannot be protected by try-except.
Typically the address is just plain bad or it is pointing at freed memory.
Arguments:
Arg1: ffffd88dea98a0ee, memory referenced.
Arg2: 0000000000000000, value 0 = read operation, 1 = write operation.
Arg3: fffff8005c9c8ea3, If non-zero, the instruction address which referenced the bad memory
	address.
Arg4: 0000000000000000, (reserved)

callstack_locals_01

yin19941005:
Adding a function for searching WCHAR in PUNICODE_STRING. This function avoid BSOD caused by the characteristics of unsigned type. Replace the similar searching implement in dokan/sys project with this new implemented function. 

This will fix #549 .

### DIFF
--- a/sys/dokan.c
+++ b/sys/dokan.c
@@ -473,12 +473,10 @@ VOID DokanNotifyReportChange0(__in PDokanFCB Fcb, __in PUNICODE_STRING FileName,
   ASSERT(Fcb != NULL);
   ASSERT(FileName != NULL);
 
-  // search the last "\"
   nameOffset = (USHORT)(FileName->Length / sizeof(WCHAR) - 1);
-  for (; FileName->Buffer[nameOffset] != L'\\'; --nameOffset)
-    ;
-  nameOffset++; // the next is the begining of filename
 
+  // search the last "\" and then calculate the Offset in bytes
+  nameOffset = (USHORT)(DokanSerachWcharinUnicodeStringWithUlong(FileName, L'\\', (ULONG)nameOffset, 1));
   nameOffset *= sizeof(WCHAR); // Offset is in bytes
 
   FsRtlNotifyFullReportChange(Fcb->Vcb->NotifySync, &Fcb->Vcb->DirNotifyList,
--- a/sys/dokan.h
+++ b/sys/dokan.h
@@ -732,4 +732,7 @@ __inline VOID DokanClearFlag(PULONG Flags, ULONG FlagBit) {
 #define DokanCCBFlagsSetBit DokanFCBFlagsSetBit
 #define DokanCCBFlagsClearBit DokanFCBFlagsClearBit
 
+ULONG DokanSerachWcharinUnicodeStringWithUlong(__in PUNICODE_STRING inputPUnicodeString, __in WCHAR targetWchar,
+	__in ULONG offsetPosition, __in int isIgnoreTargetWchar);
+
 #endif // DOKAN_H_
--- a/sys/dokan_utility.c
+++ b/sys/dokan_utility.c
@@ -1,0 +1,66 @@
+/*
+Dokan : user-mode file system library for Windows
+
+Copyright (C) 2015 - 2017 Adrien J. <liryna.stark@gmail.com> and Maxime C. <maxime@islog.com>
+Copyright (C) 2007 - 2011 Hiroki Asakawa <info@dokan-dev.net>
+
+http://dokan-dev.github.io
+
+This program is free software; you can redistribute it and/or modify it under
+the terms of the GNU Lesser General Public License as published by the Free
+Software Foundation; either version 3 of the License, or (at your option) any
+later version.
+
+This program is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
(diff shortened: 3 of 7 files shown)


sys/dokan.c
```diff
@@ -473,12 +473,10 @@ VOID DokanNotifyReportChange0(__in PDokanFCB Fcb, __in PUNICODE_STRING FileName,
...
   ASSERT(FileName != NULL);
 
-  // search the last "\"
   nameOffset = (USHORT)(FileName->Length / sizeof(WCHAR) - 1);
-  for (; FileName->Buffer[nameOffset] != L'\\'; --nameOffset)
-    ;
-  nameOffset++; // the next is the begining of filename
 
+  // search the last "\" and then calculate the Offset in bytes
+  nameOffset = (USHORT)(DokanSerachWcharinUnicodeStringWithUlong(FileName, L'\\', (ULONG)nameOffset, 1));
```
Liryna: `Search`

Liryna:
Hi @yin19941005 ,

Thank you for the pull request 👍 ! This looks good ! Just the name of the function to fix 😋

yin19941005:
Hi @Liryna ,

Sorry for the typo, I will reopen another pull request for it.

Issue #554, opened 2017-07-12
Windows 10 build 16232 become unresponsive when using winsshfs/dokany


deadalnix:
# Environment
 * Windows version: 10
 * Processor architecture: x64
 * Dokany version: 1.0.13
 * Library type (Dokany/FUSE): FUSE (I think).
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the [latest version](https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
  - [ ] I tested one of the last [snapshot](https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot) from appveyor CI

# Description

This is cross post of https://github.com/Foreveryone-cz/win-sshfs/issues/152 as I'm unsure if the problem come from winsshfs or dokany. 

The computer become unresponsive when mounting a remote filesystem using sshfs since windows 10 build 16232 . Application continue to run, and mouse still move on the screen, but it is impossible to interact with anything - click does nothing and typing on the keyboard does nothing.

I tried various version of dokany and got the same result, including the latest.

Everything works great with previous builds of windows 10. Currently download build 16237 to see if the problem persist there.

# Logs

Computer become unresponsive when the problem occurs, so not sure how to proceed here.

deadalnix:
I was able to test with windows build 16237 and it fails as well.

marinkobabic:
What are the steps to reproduce the issue? When windbg is attached using a
separate machine we should be able to discover the issue.

Does mirror work properly on those builds?

Am 12.07.2017 7:06 nachm. schrieb "deadalnix" <notifications@github.com>:

deadalnix:
Hi @marinkobabic ,

Thanks for helping.

Reading https://github.com/dokan-dev/dokany/wiki/Use-Mirror-example , it seems that mirror is supposed to be in the sample folder where dokan is installed. There is no such file there. I used the Dokan_x64.msi provided there: https://github.com/dokan-dev/dokany/releases/tag/v1.0.3

How do I get the mirror app ?

marinkobabic:
From my point of view it should be part of DokanSetupDbg.exe

[...]

PR #559, opened 2017-07-23
Resolve hang described in issue #554

benpye:
This should resolve the hang mentioned in #554 . Dokany has a few incorrect bounds checks. I am submitting this PR to get a build that I can verify the issue is resolved in.

### DIFF
--- a/sys/device.c
+++ b/sys/device.c
@@ -424,7 +424,7 @@ DiskDeviceControl(__in PDEVICE_OBJECT DeviceObject, __in PIRP Irp) {
     RtlZeroMemory(mountdevName, outputLength);
     mountdevName->NameLength = deviceName->Length;
 
-    if (sizeof(USHORT) + mountdevName->NameLength < outputLength) {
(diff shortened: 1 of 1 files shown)


benpye:
Quick test with mirror.exe and this isn't hanging any more.

Liryna:
Hi @benpye ,

Thank you for the PR !
So you were able to reproduce the issue with the commit before your fix ?

benpye:
Yeah, I can still repro with the current master. Looks like the check I linked to be fixed was the wrong one, I think CTRL+F failed me. I'm pretty sure it's the one within `IOCTL_MOUNTDEV_QUERY_DEVICE_NAME` causing the hang, as indicated by the logs and WinDbg.

Liryna:
Alright @benpye !
Thanks a lot again for looking at this fixing the issue 👍 ❤️ !

phillijw:
@benpye You're the hero we don't deserve

PR #567, opened 2017-08-29
mirror_test: Use separate dirs for each config

Rondom:
This should make testing more robust because if you only run the tests once and the directory was empty before you can be sure that no test affects the other due to files that could not be deleted because they were still in use for some reason, for example.

Using the different "names" / dirs we can distinguish between different runs as well and thus do statistics over which test cases pass/fail in which configuration.

Please double-check the code to make sure that I have not missed something.


Liryna:
All good for me @Rondom Thanks again for your contribution ! 🏆

PR #572, opened 2017-09-11
'FindClose()' is not being called under error condition

sgraves66:
Tiny issue in 'mirror'


Liryna:
Thanks you @sgraves66 for the tiny fix ❤️ !

Issue #575, opened 2017-09-21
All operations is using the mount user access token will cause Security Holes


yin19941005:
# Environment
 * Windows version: Windows 10 build 1607 (64 bit)
 * Processor architecture: AMD64
 * Dokany version: Dokan-1.0.5.1000 release
 * Library type (Dokany/FUSE): Dokany
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the [latest version](https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
  - [ ] I tested one of the last [snapshot](https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot) from appveyor CI

# Description
The following are examples which demonstrate how the security holes will occur:
### User without write permission can write file if mirror.exe started by another authorized user
Use a local admin (called `VMBB-PC6`) to mount drive with this command : `mirror.exe /r C:\dokan_temp /l Z /d /s`

1. Create a text file, e.g. testing.txt, with the same local admin account
2. Modify and Save the text file with the same local admin account
3. Set **deny write permission** to **`user01`** (another user, without admin right) with the text file (show in the image1)
4. Switch Windows User to `user01`
5. Try to modify and save the txt file with `user01`.
- **The expected behavior**:
`user01` is able to open the file but not able to write and save the text file through the mirror drive. (confirmed with physical path)
- **The actual behavior**:
`user01` is able to open, write and save the text file through the mirror drive. (**Security Issue**)

### User without permission can modify the [DACL](https://msdn.microsoft.com/en-us/library/windows/desktop/aa446597(v=vs.85).aspx) of the file if mirror.exe started by another authorized user

Use a local admin (called `VMBB-PC6`) to mount drive with this command : `mirror.exe /r C:\dokan_temp /l Z /d /s`

1. Create a text file, e.g. testing.txt, with the same local admin account
2. Modify and Save the text file with the same local admin account
3. Check the file owner, it should be the local admin account (show in the image2)
4. Switch Windows User to `user01`
5. Try to modify the text file DACL with `user01`.
- **The expected behavior**:
`user01` is able to view the DACL but not modify the DACL through the mirror drive. (confirmed with physical path)
[...]

PR #582, opened 2017-10-09
Add Impersonate Option for Security Enhancement.

yin19941005:
Add Impersonate Caller User Option for Security Enhancement. Previously implementation will always done all the operations with the mounted user access token. Adding option to use the caller access token to get the handle for completing the operations. This option requires admin right to work properly.

Can be fix the security holes mentioned in #575 when the `Impersonate Caller User Option` (require administrator right to mount) is used to mount.

### DIFF
--- a/samples/dokan_mirror/mirror.c
+++ b/samples/dokan_mirror/mirror.c
@@ -43,6 +43,7 @@ THE SOFTWARE.
 BOOL g_UseStdErr;
 BOOL g_DebugMode;
 BOOL g_HasSeSecurityPrivilege;
+BOOL g_ImpersonateCallerUser;
 
 static void DbgPrint(LPCWSTR format, ...) {
   if (g_DebugMode) {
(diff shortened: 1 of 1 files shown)


Liryna:
Hi @yin19941005 ,

This is indeed something really needed for the mirror ! Thanks a lot !

Just wonder for performance, could it be possible to only call DokanOpenRequestorToken at start and RevertToSelf when ctrl + c & exit ?

yin19941005:
Hi @Liryna ,

I afraid only call DokanOpenRequestorToken at start is not a good idea. If user with administrator right using the drive at the beginning (i.e. we impersonate his/her administrator right) and the then switch user, the same security holes on #575 will still occur. If a normal user is using the drive and switch to administrator account, the administrator will lost his/her admin right (because we are impersonating the normal user).

So, I don't think it is possible to only call DokanOpenRequestorToken at start and RevertToSelf when ctrl + c & exit.

For the performance issue, I didn't done any testing for it. However, please consider the case that if we really implement the whole security handling process instead of using impersonate method. Base on my knowledge, the DACL and SACL is require to handle them separately. DACL should required [AccessCheck](https://msdn.microsoft.com/en-us/library/windows/desktop/aa374815(v=vs.85).aspx) and SACL required [PrivilegeCheck](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379304(v=vs.85).aspx). And we also need to handle the owner ship problem (because the file will be own by the access token used to call CreateFile function).

So, I think the alternative solution required some much afford to fix this. Maybe we can reference and study the FUSE `default_permissions` option to finish these checking in kernel if you really concern about the performance.

Liryna:
OK I see your point ! and you are right.
Better to fix a security hole here 👍 

Thanks as alaways @yin19941005  !

If I am correct the `FUSE default_permissions` "kind of" already exist when DefaultGetFileSecurity is not implemented by the user, we set a default one (that can not be configured for now).
https://github.com/dokan-dev/dokany/blob/master/dokan/security.c#L29

PR #591, opened 2017-10-11
Minor fixes to README file

gkotian:
The first commit only splits long lines to 80 characters without any content changes. The second commit performs some minor typo fixes.


Liryna:
Hi @gkotian ,

Thanks you for the review and for your work ! 🎆 
Cut to 80 can be helpful for some people and anyway github does not show more 👍 It is a good idea !

Issue #587, opened 2017-10-11
Improve documentation grammar and spelling 


Rondom:
Currently [the documentation](https://dokan-dev.github.io/dokany-doc/html/) has a few grammar and spelling mistakes.

A native speaker or someone with a very good command of English should proofread it and submit changes as a pull request.
The documentation is automatically generated from [the source code](https://github.com/dokan-dev/dokany/blob/master/dokan/dokan.h) using Doxygen on every automatic build of the master branch and pushed to https://dokan-dev.github.io/dokany-doc/html/

To generate the documentation locally, one needs to install Doxygen and run `doxygen` inside the "documentations"-folder.

While one is at it, one can have a look at the homepage as well: https://dokan-dev.github.io / https://github.com/dokan-dev/dokan-dev.github.io.

klyap:
I can help!

Rondom:
We do not need any changes to the readme or mainpage.md at this point (PRs open for that already).

dokan.h still needs some improvement.

t-miller:
I'm on it!

PR #605, opened 2017-10-19
fixed typos and grammar, removed references to 'you'

t-miller:
Addresses #587.

Note that the intention of line 743 was not clear, so that sentence was not fully fixed.


Liryna:
That's an awesome review @t-miller !!!

Very kind of you for taking the time to review the dokan.h ! Can only make it more comfortable to read it 👍

PR #609, opened 2017-10-23
Add FUSE read-only option

leochou0729:


Liryna:
Hi @leochou0729 ,

This seems to be alright 👍  Thanks a lot for the PR ! ❤️

PR #614, opened 2017-10-30
FUSE: cross-compile 32-bit Cygwin DLL from 64-bit

Rondom:
Cygwin64 can compile binaries for Cygwin32. Install the corresponding
cross-compiler and use it for compiling dokan_fuse.

This should save quite a bit of time spent downloading 32-bit Cygwin
updates on AppVeyor. It also makes live easier, because there is no
longer any need to install the 32-bit Cygwin in parallel to the 64-bit
one on local machines.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -100,12 +100,11 @@
 
           function updateCygwin($cygwinexe, $installFolder, $cacheFolder) {
             Write-Host "Update Cygwin: $cygwinexe"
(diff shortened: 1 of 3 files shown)


Rondom:
Todo before merge: test that binary from AppVeyor runs

Liryna:
Hi @Rondom ,

Yes, this would make the build more fast 👍 !

There should be no issue here: gcc-g++ for x86 also use i686-pc-cygwin-c++.exe
https://cygwin.com/cgi-bin2/package-cat.cgi?file=x86%2Fgcc-g%2B%2B%2Fgcc-g%2B%2B-6.4.0-1&grep=gcc-g%2B%2B

Both packages are probably the same but one has a link to usr/bin/c++.exe directly.

This can be merge if you agree with me

Rondom:
Tested. Can be merged!

Liryna:
Thank you @Rondom ! ❤️

Issue #653, opened 2018-01-23
2nd UNC mount shows as "Disconnected Network Drive"


RustyHacksaw:
# Environment
 * Windows version: Win 7, Win 10
 * Processor architecture: x86, x64
 * Dokany version: 1.1.0.2000
 * Library type (Dokany/FUSE): Dokany
 
# Check List
  - [ X ] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [ X ] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [ X ] I can always reproduce the issue with the provided description below.
  - [ X ] I have updated Dokany to the [latest version](https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.

# Description
Using Mirror in two command prompts, mount two drives that both use UNC paths.  The first mounts fine, the second ( and subsequent ) mount as Disconnected Network Drive.  Both mounts are navigable fine, but the display in the volume label is obviously wrong.

mirror /r c:\users /l M: /n /u \myfs\myfs1
mirror /r c:\users /l N: /n /u \myfs\myfs2
 
I made sure that the dokannp1.dll was copied to the windows/system32 and that dokanctl /i n was executed before trying to mount the two drives.

RustyHacksaw:
Ah, found it, the recent addition of the SessionId information to the driver _DOKAN_CONTROL structure wasn't matched in the dokan1.dll's version of that same structure, it was missing the ULONG, hence shifting the second and subsequent structures off by 4 bytes ... I'll submit a pull request for this asap.

Liryna:
Hi @RustyHacksaw ,

Thank you for your report 👍 
And specially thanks for finding the reason of it !

Your fix is correct but I remade it by moving the struct in the sys/public.h. This header is made to share struct and avoid such issue between library and kernel.
https://github.com/dokan-dev/dokany/commit/230496e9606c2eae8b18e7566add2997a906c834
(You are the author of the commit, hope that's alright for you)

RustyHacksaw:
Hi Liryna,

Ah, okay, that's cool, I wasn't sure if there was a place for the structure.  That makes much more sense to have it in one place. =)

.r.
________________________________

PR #655, opened 2018-01-25
Add the missing session id to the user space dokan1.dll

RustyHacksaw:
This fixes bug #653 by keeping the two structures in sync.


Liryna:
https://github.com/dokan-dev/dokany/issues/653#issuecomment-360982149

PR #677, opened 2018-04-08
Undo commit bfccd62599ef80899929cbbdf7b8f0fc78f8f07b

Daniel-Abrecht:
Reverting commit bfccd62599ef80899929cbbdf7b8f0fc78f8f07b because it caused a crash in my fuse-nfs builds,
probably because the buffer size given to strncpy was wrong.
Why was this changed in the first place, the additional bound check
seams unnecessary, and it wasn't made sure the string would get null
terminated if it would ever somehow prevent an overflow. The only
way I see this could happen would be if the "opt" argument was changed
by a different thread in this tiny moment between strlen and strcpy,
if that's even possible, and even then, why would anyone ever do that,
an attacker would already have control over the program at that stage.

### DIFF
--- a/dokan_fuse/src/fuse_opt.c
+++ b/dokan_fuse/src/fuse_opt.c
@@ -103,11 +103,12 @@ int fuse_opt_add_opt(char **opts, const char *opt)
         newopts = STRDUP(opt);
     else {
         size_t oldlen = strlen(*opts);
-        size_t newlen = oldlen + 1 + strlen(opt) + 1;
-        newopts = (char *)realloc(*opts, newlen);
+        size_t optlen = strlen(opt);
+        newopts = (char *)realloc(*opts, oldlen + 1 + optlen + 1);
         if (newopts) {
             newopts[oldlen] = ',';
-            strncpy(newopts + oldlen + 1, opt, newlen);
+            strncpy(newopts + oldlen + 1, opt, optlen);
+            newopts[oldlen + 1 + optlen] = '\0';
         }
     }
     if (!newopts)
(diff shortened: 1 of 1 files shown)


Liryna:
Hi @Daniel-Abrecht ,

Thank you for your contribution.

Using strcpy is generally not recommended (code analyzer tools). I understand that here it is safe to use it but since was not hurting to change it I did. I probably failed in the len calculus, this need to be reviewed. Do you think, you could look at which one has an issue ?

Daniel-Abrecht:
Both length calculations are wrong. You passed the length of the whole buffer to strncpy, but you passed the buffer with an offset. You would have to substract that offset from the length of the buffer to get the remaining length of the buffer starting from that offset. But I think it's easier and safer to just use the length of the new option/param and add the nullbyte manually. 

Here is a commit for this:
https://github.com/Daniel-Abrecht/dokany/tree/fix_overflow_bfccd62

I'm never quiet sure how I should add such changes, if I add it to this pull request it would leaf an unnecessary commit in the git history, but if I open another pull request the discussion get's split into two parts. Can github handle a force push to a commit used to start a pull request? But then the other commit would be lost...

Rondom:
GitHub handles force pushes to pull requests fine and will at least nowadays keep the history as well. So one can still see the old commit for reference.

Daniel-Abrecht:
Ok, I've now made a force push. The new patch just corrects the length calculation and adds the null byte to terminate the strings explicitly. The previous commit seams to have been lost due to that though, it seams github is still not able to retain them if it's the first commit of a pull request.

Liryna:
Hi @Daniel-Abrecht ,

Looks good to me 👍 
Thank you for the report again and for providing a quick fix in the same time 🥇 !

PR #695, opened 2018-06-04
If a file is opened with DELETE_ON_CLOSE, then this file will disappe…

InspirationHello:
If one file is opened with DELETE_ON_CLOSE, then this file will disappear in Explorer per CREATE, such as the office's temp file.The Write modification flags should be stored in fileObject->flags.

Fixes # .
The office's temporary files in explorer are looming
### Checklist

- The Write modification flags should be stored in fileObject->flags.
- The DELETE_ON_CLOSE flag should be stored in ccb.


Liryna:
Hi @InspirationHello ,

Thank you for your pull request.
Can you give some context on the issue you faced and how to reproduce it ?
The files were blinking because of the notification happening too often ?

Have you tried to only call `DokanNotifyReportChange(fcb, FILE_NOTIFY_CHANGE_DIR_NAME` when ccb has the flag ? We should probably only notify when it is the request that holds the flag.

`OplockReadBreakItoIITest` test seems to fail with this change.
https://ci.appveyor.com/project/Maxhy/dokany/build/1.1.0-431/job/fg9xm7q4c2gr1ect?fullLog=true#L1055

InspirationHello:
The write change because: Try use office open a pptx/docx/... file, then click the tmp file in the explorer, then the temp file disappeared。
The write change because: write notify cause ReadDirectoryChanges return many times and fastfat also use fileObject to save write modification flag.

Liryna:
@InspirationHello have you been able to test using ccb for notification ?
Sorry have been quite busy this day 😢

Liryna:
Hi @InspirationHello ,

I have include one of your changes in the master https://github.com/dokan-dev/dokany/commit/29180c482f124bc6bbe9d9b14aa8704b007fe42f

InspirationHello:
I'm so sorry for the delay.
blinking

 call `DokanNotifyReportChange(fcb, FILE_NOTIFY_CHANGE_DIR_NAME` when `ccb` has the flag is OK.

InspirationHello:
in mt, `fcb->ShareAccess` is `not thread safe` may cause some file's `ShareAccess reference count` Incorrect, so cause `DokanCheckShareAccess` return not expected status,  until the `fcb's reference count` is zero.

Liryna:
https://github.com/dokan-dev/dokany/issues/769

PR #713, opened 2018-07-05
Fix: file name are uppercase when reparse point is used?

### LINKED ISSUE #293: Why file name are uppercase when reparse point is used ?
When reparse point is enabled (mount in a empty folder), all request seems to use the short file name.
We should take a look to see if it is the kernel that does not provide the correct name or it is a dokan issue.

```
> mirror.exe /r C:\Users /l C:\test /d /s
###Create 1312
CreateFile : C:\Users\LIRYNA\DESKTOP\DEBUG
  AccountName: liryna, DomainName: DESKTOP-T9B1IFA
        ShareMode = 0x7
        FILE_SHARE_READ
        FILE_SHARE_WRITE
        FILE_SHARE_DELETE
        AccessMode = 0x80
        FILE_READ_ATTRIBUTES
        FlagsAndAttributes = 0x200000
        FILE_FLAG_OPEN_REPARSE_POINT
        OPEN_EXISTING

CreateFile status = 0 - lastError = 0
###GetFileInfo 1312
GetFileInfo : C:\Users\LIRYNA\DESKTOP\DEBUG
        GetFileInformationByHandle success, file size = 4096

        result =  0
        FileBasicInformation
        DispatchQueryInformation result =  0
###Cleanup 1312
Cleanup: C:\Users\LIRYNA\DESKTOP\DEBUG

###Close 1312
Close: C:\Users\LIRYNA\DESKTOP\DEBUG

###Create 1313
CreateFile : C:\Users\LIRYNA
```

linyuan870:
Fixes #293  .

### Checklist

 Changes proposed in this pull request:

According to the information provided by the last one of #293.
https://www.osronline.com/showthread.cfm?link=287522
After making some changes to the code in the article, it successfully compiles and solves the problem. The problem of capitalization is solved, everything becomes normal; I have no problem testing on WIN10, I hope you can compile a driver as soon as possible, I need to use it recently.

The main change is to add a function that converts the name and then call the conversion function at the entry.

### DIFF
--- a/sys/dispatch.c
+++ b/sys/dispatch.c
@@ -21,11 +21,88 @@ with this program. If not, see <http://www.gnu.org/licenses/>.
 
 #include "dokan.h"
 
+#pragma warning(disable:4214)
+struct SYMLINK_ECP_CONTEXT {
+  USHORT UnparsedNameLength;
+  union {
+    USHORT Flags;
+    struct {
+      USHORT MountPoint : 1;
+    }MountPoint;
+  }FlagsMountPoint;
+  USHORT DeviceNameLength;
+  USHORT Zero;
+  struct SYMLINK_ECP_CONTEXT *Reparsed;
+  UNICODE_STRING Name;
+};
+#pragma warning(default:4214)
+
+void RevertFileName(PIRP Irp) {
+  RTL_OSVERSIONINFOW VersionInformation = {0};
+  VersionInformation.dwOSVersionInfoSize = sizeof(RTL_OSVERSIONINFOW);
+  RtlGetVersion(&VersionInformation);  
+  if (VersionInformation.dwMajorVersion > 10 ||
+      (VersionInformation.dwMajorVersion ==10 &&
+       VersionInformation.dwMinorVersion > 0) ||
+      (VersionInformation.dwMajorVersion == 10 &&
+       VersionInformation.dwMinorVersion == 0 &&
+       VersionInformation.dwBuildNumber >= 17134))
+    return;
+  
+  PECP_LIST EcpList;
+  struct SYMLINK_ECP_CONTEXT *EcpContext;
+  //IopSymlinkECPGuid "73d5118a-88ba-439f-92f4-46d38952d250";
+  GUID IopSymlinkECPGuid;
+  IopSymlinkECPGuid.Data1 = 0x73d5118a;
+  IopSymlinkECPGuid.Data2 = 0x88ba;
+  IopSymlinkECPGuid.Data3 = 0x439f;
+  IopSymlinkECPGuid.Data4[0] = 0x92;
+  IopSymlinkECPGuid.Data4[1] = 0xf4;
+  IopSymlinkECPGuid.Data4[2] = 0x46;
+  IopSymlinkECPGuid.Data4[3] = 0xd3;
+  IopSymlinkECPGuid.Data4[4] = 0x89;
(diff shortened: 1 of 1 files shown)


Rondom:
Under which Windows version have you tested this?

linyuan870:
@Rondom win 10

Rondom:
What exact version (output of command ´ver´ on command-prompt).

linyuan870:
@Rondom 
Microsoft Windows [Version 10.0.16299.431]

Liryna:
Hi @linyuan870 ,

The code is fixing the behavior 👍 Thank you
There is just a couple of warnings that need to be fixed https://ci.appveyor.com/project/Maxhy/dokany/build/1.1.0-437/job/a6063i7j3anlyi9k#L7900 and also make the function only called when repase is used.
More over, also handle the return bool from `RevertFileName` in case of failure.

Rondom:
Maybe we can test with latest Windows 10. If I understand it correctly, it should work without this patch on latest update. Maybe we can make it conditional that the code is only run on older Windows versions.

Liryna:
Did not update my windows for now to test it. If this is working, this function should be conditioned on the windows version also yes good point 👍

linyuan870:
@Liryna sorry for the late reply,
1. The warning is eliminated;
2. If
       0 <= FsRtlFindExtraCreateParameter(EcpList, &IopSymlinkECPGuid,
           (void **)&EcpContext, 0)
   Return true, that is, re-parse point; return false, that is, not re-parse point, the code will not be executed later; I think this time is not expensive; I don't know how to judge whether it is re-parsing point in the driver, you know Is there any simple way to judge?
3. I changed the return type of the conversion function to void, because this function, if it is a reparse point, converts the name accordingly; if not, it does not convert.

Rondom:
For the record, I just checked Dokan 1.1.0.2000 (i.e. without this patch) on Win 10 1803 and the filenames ended up being lowercase.
Thus I suggest to make it conditional for versions < 1803

linyuan870:
@Liryna @Rondom 
I made the windows version judgment at the beginning of the RevertFileName function. If the version >=win10 1803, I immediately returned. I did not find a way to use the macro definition to determine the specific system version of windows such as win10 1803; I found a RtlGetVersion function to get system information. 
Don't know if this change is OK?

Liryna:
Hi @linyuan870 ,

Thank you for the quick changes 👍 This seems to be the good way to do it.
You should just set dwOSVersionInfoSize from the struct before calling the function.
See https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/content/wdm/ns-wdm-_osversioninfow

Rondom:
@linyuan870 While you are at it (squashing all commits into one commit) you can apply the diff suggested by Appveyor here:
https://ci.appveyor.com/project/Maxhy/dokany/build/1.2.0-457/job/l7f373swao7mmige/messages
https://ci.appveyor.com/api/buildjobs/l7f373swao7mmige/artifacts/clang_format_267c3bd_into_f538285.diff

linyuan870:
@Liryna @Rondom OK

linyuan870:
@Liryna Hi,
Have you conducted the final test? I have squashing all commits into one commit, is it squash like this?

Liryna:
Hi @linyuan870 ,

I have been able to do the test and found no issue to it. Names are correctly converted thank you 👍

Issue #627, opened 2017-11-29
Support physical disk mirror in mirror example project


dwalkes:
Thanks for creating and sharing this very useful project!

I’m a software developer/consultant and I have a client who is interested in using this project in a new development.  They need to have access to physical drives for their application, therefore need capabilities like were discussed previously in https://github.com/dokan-dev/dokany/issues/564.  They will ultimately build dokan support into an application which is capable of decrypting their physical disk devices.  This part will be specific to their application.  However I suspect others might find the ability to interact with physical disks useful.

As a first step toward this goal, I’ve made a modification to the mirror example in [this branch](https://github.com/Trellis-Logic/dokany/tree/mirror_physical_dev_support) which can export a physical disk to a raw device file mirror on the filesystem using Dokan.  To use, invoke like this:
```
mirror.exe /P \\.\PhysicalDriveX /l "C:\temp\"
```
Where X is the physical drive number as found from disk management and C:\temp is a folder on the filesystem where a MIRROR1 raw device file will be placed.

You can then use [ImDisk](https://sourceforge.net/projects/imdisk-toolkit/) to turn this raw mount point into a mirrored physical device:
Select File->Mount New Virtual Disk
Select the C:\temp\MIRROR1 file as the Image File
Check box for “Read-only media”

There are still more general disk features needed by the client which could also be implemented in this mirror example, for instance:
1. Add write support.
2. Ideally, include features of ImDisk and support for exporting directly as a physical device to avoid including the ImDisk application and associated dependencies with their project.

My questions are:
1. Are the changes I’ve made to the mirror demo in [this branch](https://github.com/Trellis-Logic/dokany/tree/mirror_physical_dev_support) to support physical devices useful and would you like me to prepare a pull request?
2. Would you be interested in further enhancements to the mirror demo to support write of physical disks?
3. Would you like to incorporate changes to support mirroring physical disks directly, basically extending the functionality to remove dependency on ImDisk?  I haven’t looked into the details about what it would take to support this yet but I assume ImDisk could be used as an example.  Has anyone already looked into this, for instance when considering https://github.com/dokan-dev/dokany/issues/564?
4. Are any Dokan contributors interested in contract work for this client to support development of additional features for physical disk support?  If so, is [bounty source](https://www.bountysource.com/) the best way to request this for specific features?

Liryna:
Hi @dwalkes ,

This is an amazing work you did here ! 🥇 
[...]

PR #746, opened 2018-10-22
Feature/vhdx attach supporting

AGritsevich:

This merge request exist like result of implementing issue https://github.com/dokan-dev/dokany/issues/627.
There added full logic to creating VHD(X) disk images on file system with possibility to use auto attaching them in Windows file system.
It can be used for hardware encrypted disks that after sign in operation need show if system like usual disk. It's useful feature for such cases. For now, attached disk has READ ONLY access. Access for writing planing add in near future.
Solution was tested on Windows 10, Windows SDK 10.0.17134.0 and 10.0.16299.0.
See example of use:

```
mirror.exe /P \\.\PhysicalDrive3 /l C:\testMirror /e /x
/P - drive name
/l - folder for mirroring
/e - auto attach to file system as virtual drive
/x - make VHDX disk
```

### Checklist


Liryna:
Hi @AGritsevich ,

This is a very interesting work ! 
The issue I have here is that your contribution does not seem to be part of the dokan project but more like a third part FS using dokan.

You should probably create another repository outside dokan to make it public.

dwalkes:
Hi @Liryna ,

Thanks for the response.

I have one question about this comment:

> The issue I have here is that your contribution does not seem to be part of the dokan project but more like a third part FS using dokan.

Going back to your previous comments in #627:

> Usually external projects using dokan are listed on the dokan front page: https://dokan-dev.github.io/
Here I totally understand that it is a great example of implementation in another layer and could be provided as a "simple" example to physical disk mirror.

> The question of adding it in dokan samples repository (when it will be ready) is about maintaining it for me. I would need some help to maintain it, and this, for an undefined time. It should not be a day to day maintains but still need to follow to changes in dokan. @dwalkes are you willing to be an official maintainer of it if integrated in dokan repository ?

Are you no longer interested in the idea of incorporating a project like this into the samples repository?  Or is there a way @AGritsevich could transform his PR into something we could use there?

Thanks
Dan

Liryna:
Closing for inactivity

PR #772, opened 2019-01-15
sys - The szMountPoint->length should be the byte length.

wjcsharp:
During reviewing the security fix for buffer overflow issue. I found the fix may cause other issue. The szMountPoint->length is used as string length but not byte length.

### Checklist

- Div sizeof(WCHAR) before comparing with MAX_PATH


Liryna:
Hi @wjcsharp ,

Indeed this is a mistake ! 👍
Thank you for the PR and for your work that is very appreciated ! 🥇

PR #775, opened 2019-01-18
expose allocation unit size and sector size to FUSE

akkishore:
Add allocation unit size and sector size as options in dokany fuse wrapper


Liryna:
Hi @akkishore ,

Thank you for the contribution. This looks good. Even if the default value set here is 0.
Probably it is better and the dokan value will be set for out of range reason.
https://github.com/dokan-dev/dokany/blob/5d5c7bf9967ada0189127f49028962ed5852e951/dokan/dokan.c#L160-L168

PR #778, opened 2019-02-03
Increase max thread count from 15 to 63

hans-olav:
### Change proposed
The limit of 15 dispatcher threads reduces throughput when the client is employing high parallelism, for example by running `robocopy /mt:50`. I have tested with up to 63 threads with significant stress and everything seems to be running fine. For values 64 or higher, I ran into trouble, so I stopped at 63 for now.

Also note that the [Google Drive fork](https://github.com/google/google-drive-dokany/blob/master/src/dokan/dokanc.h#L52) of Dokan has this increased to 50.

Changing this limit does not change the default - it only allows code to specify something higher than 15.

### Checklist

- [X] My code follows the style guidelines of this project
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the existing documentation
- [X] I have updated the change log (Add/Change/Fix)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+### Changed
+- Increase DOKAN_MAX_THREAD from 15 to 63 for better performance.
 
 ## [1.2.1.2000] - 2018-12-21
 ### Added
--- a/dokan/dokanc.h
+++ b/dokan/dokanc.h
@@ -44,7 +44,7 @@ extern "C" {
 
 #define DOKAN_KEEPALIVE_TIME 3000 // in miliseconds
 
-#define DOKAN_MAX_THREAD 15
+#define DOKAN_MAX_THREAD 63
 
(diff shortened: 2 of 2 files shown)


Liryna:
@hans-olav Hi,

Thank you for the pull request. I can understand that this thread limit can be small depending on the user FS needs and cpu resources he has.

Have you looked at the async branch dokan? This should be much interesting for you than growing the thread limit.

hans-olav:
Is the async branch stable yet? Stability is more important than perf :) I would like to stay with mainline I think.

Is there a reason for only 15? Even Google Drive is running with much more.

- Hans
(from Phone)

On Feb 3, 2019, at 14:18, Liryna <notifications@github.com<mailto:notifications@github.com>> wrote:


@hans-olav<https://github.com/hans-olav> Hi,

Thank you for the pull request. I can understand that this thread limit can be small depending on the user FS needs and cpu resources he has.

Have you looked at the async branch dokan? This should be much interesting for you than growing the thread limit.

Liryna:
Async branch is not fully stable for now I have to say 👍 Still looking for people willing to contribut on it.

15 is a legacy value, we can increase it ofc. I just cannot say what is the "right" value. You made your test on what type of hardware ?

hans-olav:
Awesome, thanks!

Issue #782, opened 2019-02-26
Upgrading from 1.0.1.1000 to 1.2.1.2000


andreas-eriksson:
I recently upgraded my dokany version and have been having some problems since then. I created a filesystem in .NET by starting from the mirror example for dokan-dotnet. I have added the maximumComponentLength parameter to the GetVolumeInformation method. My filesystem is not very complex and only allows read access.

I now sometimes have problems with the filesystem being unmounted before I quit my program. This results in me not being able to find/read the files anymore. I would suspect that this is some sort of timeout that causes the unmount. It seems to happen very sporadically and does not always occur. I have debugged my code and there are no exceptions of any kind. I don't really know where to start my investigation and would be helpful if someone can point me in a general direction. Are there any more breaking changes between the two versions?

Liryna:
@andreas-eriksson Hi,

Could you give the full debug logs of the app and specially when it stops?

andreas-eriksson:
My app mounts several folders and one of them was somehow unmounted at 2019-02-27 08:57:48.445. The log is then silent for almost a minute even though I know that file read attempts was made.

https://www.dropbox.com/s/e1nsdx2x58ozovl/LogFile.txt?dl=0

I'm not sure if the log will actually tell you anything useful. I have tried using the .net mirror sample to read the same files and that works every time.

Liryna:
Could this be the same to https://github.com/dokan-dev/dokany/issues/779 ?

Liryna:
Will continue here #779

Issue #779, opened 2019-02-04
Mirror.exe wont stay running


jjacobsfs:
Environment
 * Windows version:  Windows 10 Pro, v10.0.17763
 * Processor architecture: x64
 * Dokany version: 1.2.1.2000
 * Library type (Dokany/FUSE): Dokany
  * Dokany command:  "C:\Dokan 1.2.1.2000\mirror.exe" /r \\nas\projects /l p

Description
Using the command above, mirror.exe will only stay running for about 5 seconds and then it shuts down.  I am not sure if this is a bug or not.  Maybe its just user error but I didn't see a forum for dokany anywhere to post this.  Thanks and I appreciate the help.

Liryna:
Hi @jjacobsfs ,

Could you provide the debug logs of the mirror ? https://github.com/dokan-dev/dokany/wiki/How-to-Debug-Dokan#mirrorfs

jjacobsfs:
Hello.  See attached.  Thanks.

Liryna:
Thank you @jjacobsfs logs seems to be truncated, possible that the app is killed ?

jjacobsfs:
Hello.  I see what you're saying but that is all I am getting in the command prompt.  Any ideas?

Thanks

Liryna:
@jjacobsfs Any anti-virus or third part that could do this?

jjacobsfs:
I shut off antivirus and any other third party apps.

I have a mapped network drive for that same network share.  Would that cause issues?  As a test, I created a new network share that was not mapped and mirror.exe ran longer than before (but it was still eventually killed).

I will keep troubleshooting and see what I can come up with.

marinkobabic:
Please try to map the windows drive and to use mapped drive as source for
the mirror

jjacobsfs:
I tried mapping the mapped drive.  It works longer but then ends up crashing Windows.  I haven't seen it do it but every time I leave mirror running and come back after several hours, Windows has restarted.

Liryna:
@jjacobsfs we will need more informations to reproduce this or logs to see what is happening

marinkobabic:
Almost in all cases this is related to keep alive but logs are needed.

taruti:
I think some of our users are also stumbling upon something similar, don't have anything enough info yet.

jjacobsfs:
[...]

PR #784, opened 2019-03-04
Keepalive activation race condition in Dokan DLL

### LINKED ISSUE #776: Keepalive activation race condition in Dokan DLL
# Environment
 * Windows version: Windows 10 Pro 
 * Processor architecture: intel x64
 * Dokany version: 1.2.1.2000
 * Library type (Dokany/FUSE): Dokany DLL
 
# Check List

# Description
There seems to be a race condition in the `DokanMain` function of the DLL (project: `dokan`, file: `dokan.c`): If the keepalive activation `DeviceIoControl(...FSCTL_ACTIVATE_KEEPALIVE...)` runs before a worker thread performs `DeviceIoControl(...IOCTL_EVENT_WAIT...)` (in `DokanLoop`) then the keepalive activation fails. You can reproduce by adding a sleep in the beginning of `DokanLoop` or not starting the threads altogether. The race condition doesn't happen in debug mode because the debugging print after mounting takes enough time for the worker thread to call `DeviceIoControl`.

On a separate note, this comment exists in `DokanMain`:
```
    // We don't consider this a fatal error because the keepalive handle is only
    // needed for abnormal termination cases anyway.
```
But it seems that if the keepalive activation fails then `DeviceIoControl(...IOCTL_EVENT_WAIT...)` in the worker thread(s) will time out (during inactivity) returning `ERROR_OPERATION_ABORTED` (995) and stopping the worker thread. I don't know if this is the correct behavior, but either the keepalive activation must be made obligatory or the worker thread should be able to recover from this error.

TrabacchinLuigi:
Fixes #776 #782 #779 .

### Checklist

- Avoid unpredictable unmount of drives, by ensuring the KeepAlive_handle is valid

All credits goes to @Hasan-Jawaheri for the code changes i just made the pull request and some tests to speed up things

### DIFF
--- a/dokan/dokan.c
+++ b/dokan/dokan.c
@@ -278,22 +278,27 @@ int DOKANAPI DokanMain(PDOKAN_OPTIONS DokanOptions,
   }
 
   wchar_t keepalive_path[128];
+  HANDLE keepalive_handle;
+  BOOL keepalive_active = FALSE;
   StringCbPrintfW(keepalive_path, sizeof(keepalive_path), L"\\\\?%s%s",
                   instance->DeviceName, DOKAN_KEEPALIVE_FILE_NAME);
-  HANDLE keepalive_handle =
-      CreateFile(keepalive_path, GENERIC_READ, 0, NULL, OPEN_EXISTING, 0, NULL);
-  if (keepalive_handle == INVALID_HANDLE_VALUE) {
-    // We don't consider this a fatal error because the keepalive handle is only
-    // needed for abnormal termination cases anyway.
-    DbgPrintW(L"Failed to open keepalive file: %s\n", keepalive_path);
-  }
+  while (!keepalive_active) {
+    keepalive_handle = CreateFile(keepalive_path, GENERIC_READ, 0, NULL,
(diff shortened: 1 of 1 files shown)


Rondom:
You can give credit in the commit itself using git commit --author="name <email>"

TrabacchinLuigi:
Argh didn't knew that, I'll try to redo it, but I have small time today...

TrabacchinLuigi:
I was wandering: can this moved to a method and be tested somehow?

Liryna:
Hi @TrabacchinLuigi ,

I understand this fix, but probably we should look deeper into this and understanding what’s happening and apply a proper fix.

TrabacchinLuigi:
IMHO this fix solve a nasty situation in a released driver with no significant downsides, I would rather merge this, test, release, and finally look deeper

taruti:
Having a release with some fix for this would be very nice.

Liryna:
Please see https://github.com/dokan-dev/dokany/issues/776#issuecomment-470262303
I will try to make a release with it after the feedback as fast as I can 👍

Issue #665, opened 2018-02-25
CI - Fail build Coverity 


Liryna:
Coverity is currently under maintenance since 20 Feb.
https://scan.coverity.com/builds
https://twitter.com/CoverityScan/status/966068113941979136

Will need to follow this and probably update when we get more news (probably they move fully to Synopsys)

Liryna:
Coverity disabled for now https://github.com/dokan-dev/dokany/commit/b9b8981e6c838e216f8421999f4efaebfa622cde 
Still no update

Rondom:
Coverity is back, but everyone is advised to change their credentials (I guess besides the password it includes the credentials stored in "secure environment variables" in AppVeyor)

Liryna:
Hi @Rondom ,

Yes coverity is back but since our build fail with http error 500 BUT the build is correctly sent and analyzed.
https://ci.appveyor.com/project/Maxhy/dokany/build/1.1.0-407/job/byorkqn2pit61q2d#L524
I have updated the api key as you said and disabled the return check value for now.
https://github.com/dokan-dev/dokany/commit/3afd8dcb8def766885d33939e0f992c57cae6033

I also try to contact coverity support for the reason

Liryna:
Have created an issue on coveritypublisher that we are using
https://github.com/csMACnz/coveritypublisher/issues/16

Liryna:
Coverity was removed

PR #797, opened 2019-05-05
Add build and vars scripts to support building with sonarqube analysis

kyanha:
### Checklist

- [X] I have commented my code, particularly in hard-to-understand areas
- [] I have made corresponding changes to the existing documentation

- Add a new script, based on build.bat, which will run the Sonar Scanner for MSBuild and the Sonar MSBuild Build Wrapper against the build.  I understand that the project itself has a Travis Continuous Integration system set up for it; however, this is for the benefit of individual developers who want to run SonarQube/SonarCloud analysis against their own changes.
- Modify the vswhere.exe command to reverse-sort the Visual Studio versions installed on the system.  I currently have both VS2017 and VS2019 installed; I found out the hard way that this altered the default output of vswhere.exe.  (I would do a simple version specification, but the version specification language available for vswhere breaks the FOR processing.  See https://developercommunity.visualstudio.com/content/problem/556659/vswhereexe-documented-version-range-syntax-is-mang.html for the bug I filed on it.)
- Reorder the build, so that the non-driver builds (both Release and Debug) are in one section and the driver builds (both Release and Debug) are in another.  Also, so that x64 is built before Win32.  This is to work around a SonarQube issue where only the first build is actually processed; this caused the sys/ directory to not show up in the analysis, because it isn't built for the non-OS-version-specific builds.  Building x64 before Win32 is to make it so that the code build that's running on my system is the code that's analyzed, and can be undone if you'd prefer.

If documentation needs to be added to explain what needs to be changed to get it to work (basically, sonarvars.bat needs to be modified to plug in the correct values available from e.g. sonarcloud.io), I'll happily write it.  Also, sonarvars.bat is a separate file so that it can perhaps be added to .gitignore so that secrets don't inadvertently get put in public repositories, while allowing the main logic of the script to be updated as appropriate.

### DIFF
--- a/sonarbuild.bat
+++ b/sonarbuild.bat
@@ -1,0 +1,150 @@
+@echo off
+setlocal ENABLEDELAYEDEXPANSION
+REM Build release to prepare for packaging
+
+REM Make sure MSBUILD is available
+REM (the -sort option reverses the sort order. If/when dokany moves to VS2019, this will need to be changed.  The logic will need to be revisited when a VS version later than 2019 comes out.)
+FOR /f "delims=" %%A IN (
+	'"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -property installationPath -sort'
+) DO SET "VS_PATH=%%A"
+
+SET MSBUILD_BIN_PATH="%VS_PATH%\MSBuild\15.0\Bin"
+IF NOT EXIST "%VS_PATH%" (
+	ECHO Visual C++ 2017 NOT Installed.
+	PAUSE
+	EXIT /B
+)
+
+rem This is where you should set the environment variables needed for integration with SonarQube.
+IF EXIST "sonarvars.bat" (
+	CALL sonarvars.bat
+)
+
+rem Seriously, the space after the openparen must be there to separate the tokens, and there must be no space between the end of the SET values and the closeparen. Otherwise, the trailing space
+rem gets put into the value of the environment variable being set, which causes all the string equality tests to fail.
+IF NOT "%SONARBUILD%"=="" ( SET shouldsonar=true)	
+IF NOT "%SONAR_KEY%"==""  ( SET shouldsonar=true)
+IF NOT "%shouldsonar%"=="true" goto skipsonar 
+	ECHO Building using SonarBuild. Environment variables which must be set are:
+	ECHO SONAR_KEY:				The parameter to /k: for user-space components (%SONAR_KEY%)
+	ECHO SONAR_KEY_SYS:			The parameter to /k: for kernel-mode components (%SONAR_KEY_SYS%)
+	ECHO SONAR_LOGIN:			The parameter to /d:sonar.login= (%SONAR_LOGIN%)
+	ECHO SONAR_ORGANIZATION:	The parameter to /d:sonar.organization= (%SONAR_ORGANIZATION%)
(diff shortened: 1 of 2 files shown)


sonarbuild.bat
```diff
@@ -0,0 +1,150 @@
...
+
+
+REM Set up SonarQube kernel-mode analysis if SONAR_KEY_SYS environment variable set
+IF NOT "%BUILD_WRAPPER%"=="" (
+	IF NOT "%SONAR_KEY_SYS%"=="" (
+		%SONAR_SCANNER% begin /k:"%SONAR_KEY_SYS%" /d:sonar.login="%SONAR_LOGIN%" /d:sonar.organization="%SONAR_ORGANIZATION%" /d:sonar.host.url="%SONAR_HOST_URL%" /d:sonar.cfamily.build-wrapper-output=!SONAR_CFAMILY_BUILDWRAPPEROUTPUT!
+	)
+)
+
+%BUILD_WRAPPER% msbuild dokan.sln /p:Configuration="Win10 Release";Platform=x64 /t:Rebuild !CI_BUILD_ARG! || goto buildfailedsys
```
Rondom: I have not cross-checked this, and my batch-skills became a bit rusty over the years: Can we somehow share some code with build.bat?

kyanha: I actually wrote this to do exactly the same as build.bat if it doesn't have the SonarQube variables set, which means "if sonarvars.bat doesn't exist".  (I could rem out the SET commands in sonarvars.bat so that they wouldn't run, but it would make things more difficult if people didn't already understand that they needed to remove the REM statements.  That's why I didn't offer a patch to overwrite build.bat.)

What I did was take build.bat and add a bunch of variable setting and checking.  I also added %BUILD_WRAPPER% invocations before the 'msbuild' lines, because if you don't run it through the build wrapper then SonarScanner doesn't capture the output to figure out what it's supposed to analyze.  (if %BUILD_WRAPPER% isn't set, then it expands to an empty string, so the msbuild command becomes the first thing on the line.)

Theoretically, it might be possible to share code with build.bat.  It would require changing the order of the builds in build.bat (because of SonarScanner's ordering issue and the sys/ directory not being compiled in the non-version-specific msbuilds), and would require some fancy string matching in a FOR loop (so that only lines that start with "msbuild" would be invoked).  It would also be lot more difficult (read: I personally can't figure out a way) to separate out the non-version-specific builds for general analysis and the version-specific builds for sys/ directory analysis.  Input on how to accomplish it would be appreciated, since sys/ directory analysis is a concern.

Liryna: Hi @kyanha,
I agree with @Rondom it is a good idea you have but it share too much code with `build.bat`. We do not need to stick with batch code.
What could be done is to move `build.bat` and `dokan_wix` in scripts folder. Rewrite them in powershell to use helpers (`build_helper.ps1`) to avoid duplicate codes like vswhere / foreach build configuration and platform, .... Even create a coverity build script that use/download official tools instead of coveritypublisher https://github.com/dokan-dev/dokany/issues/665
Like that appveyor script would be smaller and only set the needed env variables for coverity / sonar like you did with `sonarvars.bat` and call our build scripts.

Liryna:
Rewrote in ps here: https://github.com/dokan-dev/dokany/commit/bd8eb71281db1051d59005015e6fbe656dc10903

Liryna:
Done with https://github.com/dokan-dev/dokany/commit/739a9645ec83f117a4fc35d82826617b1b9ed48a

Thank you @kyanha for leading this idea 👍

PR #803, opened 2019-06-01
Misc tooling + build settings updates.

woachk:
- Update the used SDK + WDK to the one from Windows 10 version 1903. 
- This also updates the Visual Studio toolset version to VS2019 (v142).
- This disables SSE ISA extensions for the Windows 7 kernel drivers & the user-space tools, for compatibility with older machines.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -31,6 +31,7 @@
     - Visual Studio 2015
 # If you suspect some changes on the AppVeyor-side breaking things,
 # uncomment below. https://www.appveyor.com/updates/
+#    - Previous Visual Studio 2019
 #    - Previous Visual Studio 2015
 #    - Previous Visual Studio 2017
   matrix:
@@ -262,7 +263,7 @@
         $env:Path = $env:Path + ";C:\Program Files (x86)\WiX Toolset v3.10;C:\Program Files (x86)\Windows Kits\8.1\bin\x64\"
         $env:CERTISSUER="DokanCA"
         $env:ADDITIONALCERT="$pwd\cert\DokanCA.cer"
-        $env:SIGNTOOL="C:\Program Files (x86)\Windows Kits\8.1\bin\x64\signtool.exe"
+        $env:SIGNTOOL="C:\Program Files (x86)\Windows Kits\10\bin\10.0.18362.0\x64\signtool.exe"
 
         cd dokan_wix
         $Platform = "Any CPU"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,7 +15,7 @@
       MSYS2_CACHE: '%DOKAN_CI_CACHE%\msys2'
       CHOCO_CACHE: '%DOKAN_CI_CACHE%\choco'
       WLK_INST_CACHE: '%DOKAN_CI_CACHE%\wlk_inst'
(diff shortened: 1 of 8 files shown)


Liryna:
Hi @woachk 

Thank you for the PR, I haven't moved my env to 2019 for now but will do soon 👍 
Just about the SSE2 / ISA I see your point just that it really disables for all other arch here. Does `NoExtensions` also do not disable everything and not just the SEE2 / ISA (never used the flag) ?
What happens if you use on an older CPU without SSE2 for now ? Should he not just escape it ?

woachk:
@Liryna SSE/SSE2/AVX are the ISA extensions disabled by NoExtensions, MMX and such are considered as the baseline, and are still enabled in that mode. By default if you don't specify the setting, the baseline for MSVC is SSE2. NoExtensions only covers the ISA extensions.
For now, Dokan just crashes on old systems without SSE2, but there's so little of them these days. (mainly Athlon XPs and Pentium !!!s).

Note that this only affects 32-bit mode of course, and that 64-bit is not affected. On 64-bit Windows, the baseline feature set is SSE2 (+CMPXCHG8B).

2019 again didn't change the VS runtime version like the move from 2015 to 2017, it's still msvcr140 and co, still recommended to update the runtime version though.

Liryna:
Thank you @woachk for all this Information 👍 
Am surprise SSE2 is not properly handled. I found an issue in VS 15.5 but it seems to have been fixed since.
https://developercommunity.visualstudio.com/content/problem/166375/avx-instructions-generated-since-155-update.html
Even if it seems to be a compiler bug. I don't think we should disable it just for some minor old CPU 32-bit.

Could you remove the disables SSE ISA extensions commit for your PR ?

For the VS runtime, I agree, I can look into it to update it.

woachk:
@Liryna Hi, that issue was even worse, because AVX is only available on Ivy Bridge onwards (and is not available at all on Pentium and Intel Atom mobile parts, the issue was just way too big to stay). 

About CPUs without SSE2, the issue only affects old Athlon XP and Pentium III machines (and earlier), with a sole exception, the Intel Galileo board with Windows IoT 8.1, which is a special case. Maybe it's just not worth it to support them.

Liryna:
@woachk I do not think it is worth supporting it yes but it is good to know !
Otherwise I would like to merge your vs 2019 / new wdk changes. Could you remove the commit around SSE ? Thank you

woachk:
Ok, done.

Liryna:
Thank you @woachk there is just two last NoExtensions in the sys project
https://github.com/dokan-dev/dokany/pull/803/files#diff-e8401a5bc872189984bafceb1cf636a5R737

woachk:
@Liryna It's done on purpose because of https://docs.microsoft.com/en-us/windows-hardware/drivers/kernel/floating-point-support-for-64-bit-drivers for Windows 7.

KeSaveExtendedProcessorState isn't used in this driver, and this with enabled ISA extensions can cause dire consequences.

woachk:
Setting removed because it turns out that it currently works fine without it (tested), might be a tad risky though.

Liryna:
Thank you @woachk 👍 for the detail and the pull request. Very appreciated as usual 🎖

PR #845, opened 2019-11-04
Log redirection

KerberX:
Adds log redirection options.

### Checklist

Currently there is no way to make Dokany log its operation to user-defined target. There is an option to use `stderr` instead of `OutputDebugStringA/W` but that might not be desirable to the library user.

I've added a struct with two function to be provided by the library user that will take formatted debug string and return a `BOOL` value which indicated in Dokany should proceed with its own debug output. The value is `TRUE` by default so in case the used didn't provide any log redirection function the library will output log itself.

### DIFF
--- a/dokan/dokan.c
+++ b/dokan/dokan.c
@@ -84,6 +84,9 @@ volatile PDokanMalloc	g_DokanMalloc = NULL;
 volatile PDokanFree		g_DokanFree = NULL;
 volatile PDokanRealloc	g_DokanRealloc = NULL;
 
+volatile PDokanDbgPrint g_PDokanDbgPrint = NULL;
+volatile PDokanDbgPrintW g_PDokanDbgPrintW = NULL;
+
 volatile LONG			g_DokanInitialized = 0;
 
 // TODO NEXT:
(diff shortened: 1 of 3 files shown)


KerberX:
Sorry, I'm idiot and couldn't find all places with `DokanInit` at once.

KerberX:
Now it has been built successfully.

Liryna:
Hi @KerberX ,

That's a good idea! Look at the function pointer, would it not be better to directly call the user function instead of our ? This way he will manage as he wants how to log it.

Like DokanDbgPrint would create the debug output and then call PDokanDbgPrint -> return Or follow the current output path.

KerberX:
Yeah, that would be better indeed. I'll make the changes.

Could it be possible to implement such thing for v1.x? As far as I know there is no global Init/Shutdown routine that can initialize logging functions.

Liryna:
It can be possible to implement it yes at mount time.
This makes me think we should (my point of view) maybe add a param to inform from which instance this log comes from ?
Or maybe not use global function pointers but function pointer attached to the instance.
I agree this is more than this PR as it is already the case to log all instances at the same place 😃

KerberX:
> Or maybe not use global function pointers but function pointer attached to the instance.

In the same fashion as operation callbacks? I'd like that solution.

Liryna:
Yes this way but maybe not directly to the dokanoperations 😄

Liryna:
The new V2 has a new mandatory API named DokanInit.
@Karandra if you are still interested, it would be great if you could port your work by using the API to register the struct logging functions.

Karandra:
I'll see what I can do. I still need to see has been changed between that version and the current one.

PR #850, opened 2019-11-13
append to file failed with git bash echo ">>" 

nightwend:
Fixes # .

### Checklist

git bash append to file use -1 as offset. NTFS works well with it. dokany fuse reject -1 as offset. The patch load file size as offset.
dokany fuse check for share mode in open file. But it includes the share mode requested in the open file request. It blocks windows "echo >>xxx" appending command from cmd.


Liryna:
Hi @nightwend ,

Thank you for your pull request very appreciated and sorry for the delay !

You are right about write -1 offset 👍 it needs to be handled as you did
`To write to the end of file, specify both the Offset and OffsetHigh members of the OVERLAPPED structure as 0xFFFFFFFF. This is functionally equivalent to previously calling the CreateFile function to open hFile using FILE_APPEND_DATA access.`
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-writefile

PR #857, opened 2020-01-05
Add support for compiling with GCC

ravenpride:
### Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas (does not apply)
- [ ] I have made corresponding changes to the existing documentation (does not apply)

This PR adds support for cross-compiling the `dokan` subproject with GCC and mingw32-w64.
Building with mingw32-w64 eliminates the dependency to the Visual Studio C++ Runtime and allows building on a Linux build server (all our projects - even those targetting Windows - are built on Linux).

### DIFF
--- a/dokan/directory.c
+++ b/dokan/directory.c
@@ -23,8 +23,10 @@ with this program. If not, see <http://www.gnu.org/licenses/>.
 #include "fileinfo.h"
 #include "list.h"
 
+#ifdef _MSC_VER
 #if _MSC_VER < 1300 // VC6
 typedef ULONG ULONG_PTR;
+#endif
 #endif
 
 /**
--- a/dokan/dokan.rc
+++ b/dokan/dokan.rc
@@ -5,7 +5,7 @@
 //
 // Generated from the TEXTINCLUDE 2 resource.
 //
-#include "winres.h"
+#include "winresrc.h"
 
 /////////////////////////////////////////////////////////////////////////////
 #undef APSTUDIO_READONLY_SYMBOLS
--- a/dokan/dokanc.h
+++ b/dokan/dokanc.h
@@ -52,10 +52,10 @@ extern BOOL g_DebugMode;
 // DokanOptions->UseStdErr is ON?
 extern BOOL g_UseStdErr;
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || (defined(__GNUC__) && !defined(__CYGWIN__))
 
 static VOID DokanDbgPrint(LPCSTR format, ...) {
-  const char *outputString = format;    // fallback
+  const char *outputString = format; // fallback
   char *buffer = NULL;
   int length;
   va_list argp;
@@ -80,7 +80,7 @@ static VOID DokanDbgPrint(LPCSTR format, ...) {
 }
 
 static VOID DokanDbgPrintW(LPCWSTR format, ...) {
-  const WCHAR *outputString = format;   // fallback
+  const WCHAR *outputString = format; // fallback
   WCHAR *buffer = NULL;
   int length;
   va_list argp;
@@ -102,6 +102,8 @@ static VOID DokanDbgPrintW(LPCWSTR format, ...) {
   va_end(argp);
 }
 
+#if defined(_MSC_VER)
+
 #define DbgPrint(format, ...)                                                  \
   do {                                                                         \
     if (g_DebugMode) {                                                         \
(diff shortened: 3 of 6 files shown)


ravenpride:
Two things to mention:
- The PR renames same includes to lower case as mingw contains these headers in the lower-case form. It does not harm when building on Windows as file names are case-insensitive there.
- In `dokan.rc` there was `winres.h` included. Some environments do not contain `winres.h` (such as mingw), using `afxres.h` would work, but breaks builds that do not know anything about the MFC. I think a good workaround is to simply include `windows.h` in this case. This header is always available when building for Windows.

ravenpride:
I've finished and the PR is ready for review.

Thank you very much in advance!

ravenpride:
I've found a better alternative to including `windows.h` in `dokan.rc` [here](https://github.com/zufuliu/notepad2/issues/10). The last commit integrates this. Including `windows.h` has the downside that it does not set `IDC_STATIC` as `afxres.h` and `winres.h` does.

dokan/dokan.rc
```diff
@@ -5,7 +5,7 @@
 //
 // Generated from the TEXTINCLUDE 2 resource.
 //
-#include "windows.h"
+#include "rcinc.h"
```
Rondom: Why does winresrc.h not work? It works fine with dokan_fuse which is compiled both with MingW-w64 and the MS toolchain.

https://github.com/dokan-dev/dokany/blob/0c152df921c05951788e29a69cb8c8158764733f/dokan_fuse/src/dokanfuse.rc#L8

dokan/directory.c
```diff
@@ -23,9 +23,11 @@ with this program. If not, see <http://www.gnu.org/licenses/>.
 #include "fileinfo.h"
 #include "list.h"
 
+#ifdef _MSC_VER
 #if _MSC_VER < 1300 // VC6
```
Rondom: `#if defined(_MSC_VER) && _MSC_VER < 1300`

But the question is whether we really do support VC6 in 2020. If not, we might as well remove this. @Liryna What do you think?

dokan/dokanc.h
```diff
@@ -52,10 +52,10 @@ extern BOOL g_DebugMode;
 // DokanOptions->UseStdErr is ON?
 extern BOOL g_UseStdErr;
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || (defined(__GNUC__) && !defined(__CYGWIN__))
```
Rondom: I will see whether I find the time to look into this more closely at the end of the week. Maybe we can make it easier somehow.

dokan/timeout.c
```diff
@@ -64,7 +64,8 @@ BOOL DOKANAPI DokanResetTimeout(ULONG Timeout, PDOKAN_FILE_INFO FileInfo) {
 }
 
 /* Legacy KeepAlive - Remove for 2.0.0 */
-UINT WINAPI DokanKeepAlive(PDOKAN_INSTANCE DokanInstance) {
+UINT WINAPI DokanKeepAlive(PVOID instance) {
```
Rondom: Can you elaborate on the why here?

Rondom:
Good work :-)

Liryna:
@ravenpride Thank you very much for the contribution! That's an interesting one :)! 

Have a couple of things going on but be sure I will look to review/merge your PR in the next days.

Do you think there would be a sense to include those binaries in the installer?

dokan/dokan.rc
ravenpride: Hmm, `winresrc.h` didn't come to my mind. I'll give it a try - it surely is much less complex as my solution.

ravenpride: Ok, that works for me as well. Thank you for the hint!

dokan/timeout.c
ravenpride: `UINT WINAPI DokanKeepAlive(PDOKAN_INSTANCE DokanInstance)` has a body.

`UINT WINAPI DokanKeepAlive(PVOID instance)` did not have a body, only a declaration in `dokani.h`.

Some compilers seem to accept the ambiguity and map both function declarations to the same function body, but my GCC did not. The change was necessary to fix it.

ravenpride:
> @ravenpride Thank you very much for the contribution! That's an interesting one :)!
> 
> Have a couple of things going on but be sure I will look to review/merge your PR in the next days.
> 
> Do you think there would be a sense to include those binaries in the installer?

Which binaries do you mean? This PR does not introduce any new files.

dokan/timeout.c
Rondom: Hm, I am not sure why the prototype differs from the definition in the first place.

@Liryna I am wondering why it is different in the first place. Is there anything that speaks against making the definition in timeout.c take a `PDOKAN_INSTANCE` instead of `PVOID`?

ravenpride: To make it clearer, I considered to use `UINT WINAPI DokanKeepAlive(PDOKAN_INSTANCE DokanInstance)` in `dokani.h`, but that would have broken the API as the signature of the function changes.

Liryna:
@ravenpride Not yet :-D but if it sounds interesting for some people we can add gcc build of the dokan library now that your PR allows it and we can propose them on the installer.

ravenpride:
@Liryna Please notice that this PR focuses on the `dokan` subproject. I've not tested building the other subprojects as they were not needed. I've built a CMake configuration to build it, if you're interested.

ravenpride:
Should I squash the changes before the merge?

Liryna:
@ravenpride Yes please! Thanks you

ravenpride:
Ok, I've squashed the commits.

dokan/dokanc.h
Rondom: no, I won't find the time. Sorry!

Liryna:
Hi @ravenpride 

Sorry for the long delay, I should be quicker now to review PR if you have other idea pending so please do not hesitate.

Thank you again for the contributions it is always appreciated ! :trophy:

ravenpride:
@Liryna Thank you for merging :-)

Issue #734, opened 2018-09-10
fuse read after close


lijie0123:
# Environment
 * Windows version: 10.0.17134.228
 * Processor architecture: amd64
 * Dokany version: 1.2.0.1000
 * Library type (Dokany/FUSE): FUSE
 
# Check List
  - [x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
  - [x] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the [latest version](https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
  - [ ] I tested one of the last [snapshot](https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot) from appveyor CI

# Description
fuse will still read after the file is closed and cause invalid memory access.
The bug is in method ''impl_fuse_context::read_file'' in file "fusemain.cpp", line 614. While in this read loop, the file is closed at another thread, the hndl will be deleted, and cause the dangling pointer access.

# Logs

Liryna:
Hi @lijie0123 ,

Can you provide a sample to reproduce the behavior ?
Normally kernel would not request a close if there is still a read pending. Or it means the read was taking too much time and kernel has supposed the request has timeout.

lijie0123:
My log is bellow:
14984 TRACE [11328] read: ENTER: path=/hisense_demonstration_visual-DWEU.mkv,offset=335544320,size=65536 
15121 TRACE [13380] release: ENTER: path=/hisense_demonstration_visual-DWEU.mkv
15130 TRACE [13380] release: EXIT:  path=/hisense_demonstration_visual-DWEU.mkv
15199 TRACE [11328]  read: EXIT:  path=/hisense_demonstration_visual-DWEU.mkv,offset=335544320,size=65536 

When I use windows movies & TV application to open a video and drag progress bar and close the app, the fuse wrapper crashed because of access invalid memory. But with VLC, it is all OK, vlc will close after the read procedure return back.

When I update macro  MAX_READ_SIZE  to a larger number, this problem is avoided.

Reading process only take 0.2s and it is not timed out because with a larger MAX_READ_SIZE, it is not time out.

[...]

PR #858, opened 2020-01-07
fix read file exception

hefengliang:
Fixes # .
When reading data with a large delay, the IRP times out, which will cause the file object to be accidentally released.
When reading data in multiple reads, due to the dependency on the file object (hndl-> get_name ()), uncontrollable pointer errors will result.
This question has also been asked before: https://github.com/dokan-dev/dokany/issues/734

I fixed this problem, and through actual testing, I found that there would be no more file reading problems, but I do n’t know if the data returned by read can still be received by the kernel and processed correctly when the file object is released. At least my application looks normal.

### Checklist

-Compare the code, only two lines of code changed
-
-

### DIFF
--- a/dokan_fuse/src/fusemain.cpp
+++ b/dokan_fuse/src/fusemain.cpp
@@ -605,14 +605,14 @@ int impl_fuse_context::read_file(LPCWSTR /*file_name*/, LPVOID buffer,
   CHECKED(cast_from_longlong(offset, &off));
   fuse_file_info finfo(hndl->make_finfo());
 
-  std::string strName = hndl->get_name();
+  std::string file_name = hndl->get_name();
   DWORD total_read = 0;
   while (total_read < num_bytes_to_read) {
     DWORD to_read = num_bytes_to_read - total_read;
     if (to_read > MAX_READ_SIZE)
       to_read = MAX_READ_SIZE;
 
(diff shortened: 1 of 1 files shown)


hefengliang:
This is the state when the exception occurred：
image

Liryna:
Hi @hefengliang ,

I see your situation, you are getting a read / write after the object was released. This fix just avoid the crash but do not fix the behavior.
You can get a read / write after the cleanup but not after the close. Can you track when the name is released in your case ?

hefengliang:
In my application, this problem only occurs in the case of a poor network. Due to the complicated environment, please forgive me for not being able to write a demo. I exported an abnormal dmp file. Is it useful to you?
dmp is based on Dokan Library-1.3.1 \ x86.
Since there is a 10M limit here, I sent the dmp file to your email address：liryna.stark@gmail.com

Liryna:
Sorry for the delay @hefengliang Could you please rebase the commit into one for me to merge ?

hefengliang:
I use svn to develop, GIT is not commonly used, I merge as I understand

hefengliang:
@Liryna   I use svn to develop, GIT is not commonly used, I merge as I understand

Liryna:
Thank you @hefengliang for your contribution ! :+1:

PR #863, opened 2020-01-16
feat: use stbuf in dokan_fuse readdir callback

vincentdchan:
Fixes # .

### Checklist

- use stbuf params in readdir callback, which can improve performance

### DIFF
--- a/dokan_fuse/src/fusemain.cpp
+++ b/dokan_fuse/src/fusemain.cpp
@@ -297,6 +297,9 @@ int impl_fuse_context::walk_directory(void *buf, const char *name,
   if (strcmp(name, ".") == 0 || strcmp(name, "..") == 0) {// Special entries
     stat.st_mode |= S_IFDIR; // TODO: fill directory params here!!!
   }
+  else if (stbuf != nullptr) {
+    memcpy(&stat, stbuf, sizeof(FUSE_STAT));
+  }
   else if (ctx->ops_.getattr) {
     CHECKED(ctx->ops_.getattr((dirname + name).c_str(), &stat));
   }


Liryna:
Hi @vincentdchan ,

Could you explain a little more when does this improve performance for you ?

vincentdchan:
@Liryna Of course. 

When dokan_fuse call `readdir`, user's code use `filler` to fill the data for dokan, the filler function is `impl_fuse_context::walk_directory`.  The second parameter of `walk_directory` is a pointer to the `FUSE_STAT` struct, which can provide the stat information of that filling item. But unfortunately, dokan_fuse doesn't use it, it call another `get_attr` to acquire this FUSE_STAT from user's callback. Imagine a directory with thousands of files, the `readdir` callback will call thousands of `get_attr` callback, which is equivalent to thousands of disk IO. It would be very slow.

Actually, these function calls are redundant, user's callback can provide STAT struct in `readdir` callback and fill them to dokan_fuse. It would save a lot of time of IO. This PR would use the second parameter of `walk_directory` if it's not a null pointer. It can improve performance.

Rondom:
Fuse 2.x servers are not required to fill those attributes by default. What you are looking for is adding "READDIR_PLUS" support which was added to sometime later to libfuse 2.x, but has not yet been added to dokan_fuse.

vincentdchan:
@Rondom Yep, it's not required, it can be NULL and I have considered that in my code. But the parameter has been defined in the code of fuse, why not use it? It can improve performance.

zqlustc:
Hi, 
  For help!  
  I am using dokan to develop a filesystem, and I have to modify the dokan sys. Now I need to modify a few lines of code in dokan sys. But I do not know how to build it and make dokan1.sys. Could you provide a wiki for guide. Thanks very much.
Thanks
zhangqilong

Liryna:
Sorry @vincentdchan and @Temtaime this was out of my radar for a long time!

Thanks @vincentdchan again for your contribution !

PR #875, opened 2020-03-26
Allow disconnecting of network drives

phobeus:
Fixes https://github.com/dokan-dev/dokan-dotnet/issues/250

I could not figure out how to pass down the setting, but this should nonetheless be a good new default behavior in my opinion.

### Checklist

- Add the option to enable disconnecting / unmounting a network drive from explorer directly

### DIFF
--- a/samples/dokan_mirror/mirror.c
+++ b/samples/dokan_mirror/mirror.c
@@ -1549,7 +1549,8 @@ void ShowUsage() {
           "  /f User mode Lock\t\t\t\t Enable Lockfile/Unlockfile operations. Otherwise Dokan will take care of it.\n"
           "  /e Disable OpLocks\t\t\t\t Disable OpLocks kernel operations. Otherwise Dokan will take care of it.\n"
           "  /i (Timeout in Milliseconds ex. /i 30000)\t Timeout until a running operation is aborted and the device is unmounted.\n"
-          "  /z Enabled FCB GC. Might speed up on env with filter drivers (Anti-virus) slowing down the system.\n\n"
+          "  /z Enabled FCB GC. Might speed up on env with filter drivers (Anti-virus) slowing down the system.\n"
+          "  /x (network unmount)\t\t\t Allows unmounting network drive from file explorer\n\n"
           "Examples:\n"
           "\tmirror.exe /r C:\\Users /l M:\t\t\t# Mirror C:\\Users as RootDirectory into a drive of letter M:\\.\n"
           "\tmirror.exe /r C:\\Users /l C:\\mount\\dokan\t# Mirror C:\\Users as RootDirectory into NTFS folder C:\\mount\\dokan.\n"
@@ -1636,6 +1637,9 @@ int __cdecl wmain(ULONG argc, PWCHAR argv[]) {
       break;
     case L'z':
       dokanOptions.Options |= DOKAN_OPTION_ENABLE_FCB_GARBAGE_COLLECTION;
+      break;
+    case L'x':
+      dokanOptions.Options |= DOKAN_OPTION_ENABLE_UNMOUNT_NETWORK_DRIVE;
       break;
     case L'b':
       // Only work when mirroring a folder with setCaseSensitiveInfo option enabled on win10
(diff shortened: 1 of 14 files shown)


Liryna:
Thank you @phobeus for your PR !
We can try to get the mount option by adding an IOCTL that would directly return all the mount information including the options.
Like [IOCTL_EVENT_MOUNTPOINT_LIST](https://github.com/dokan-dev/dokany/blob/master/sys/public.h#L68) but it would take the Mountpoint and return a struct containing all informations we can safely return to userland (with the option that interest you)

phobeus:
Hey @Liryna I hope you are well.
Thanks for your insights. I will try to implement that and don't want to be a pest, but did you have time to look at my comments on the linked issue? We have a release coming up in 2 weeks and need an installer with this fix in it. I would just build it myself but was so far unsuccessful.
We want to use the official releases if possible but it doesn't look like any release is going to happen any time soon. Thanks so much for the time you put into this project.
I'll update this PR accordingly as soon as I find the time to do this properly.

phobeus:
@Liryna It's not looking like we'll have time to implement this properly ourselves, at least not without help.
I've tried to get this running last time using this code: https://github.com/phobeus/dokany/commit/8832f5b962293b29b1bfbc866debca60d054a008 but the setting is always given as the default value. Due to my inexperience in this area, I was hoping you could point me into the right direction?

If anyone else has any idea how to implement this setting, any help is greatly appreciated.
Another option would be to at least make this a new default, I don't think it's such a bad thing if users can dismount their drives?

Liryna:
@phobeus I unfortunately not really have time to dedicate for this feature.
But you are not far to have something good to be merged! I added some comments 👍

phobeus:
Thank you very much for the hints, I got it running now: https://github.com/dokan-dev/dokany/commit/9201a0410816b3c6fcd15eb33c66fc0477e8396b
I'm unsure about the settings though, I can see that the start flags are being set in DokanStart and then passed down as Flags to DokanEventStart where all items just have their own boolean, that's why I did the same now.

I can make another flags item that incorporates all the settings again and store that instead of the new boolean and maybe remove them from the DokanDiskControlBlock - is that what you meant? If yes, how am I able to test that I didn't break something (other than smoke testing it with mirror.exe)? Will the CI catch that?

phobeus:
To follow up on that, the event.c in the sys assembly would probably have to do that in DokanEventStart, but the options are all defined in dokan.h so I'd have to use the DOKAN_EVENT_* items to specify an options item. Would that be fine too? Or does it have to be the DOKAN_OPTION_* items?

Liryna:
@phobeus Cannot see the change for now but yes you got it. Storing the dokan options as they came from the userland in dokaneventstart and sending it in the dokan_control is the good way!

phobeus:
@Liryna I updated all references where it made sense to me. Pull request is now up to date again.

phobeus:
@Liryna if there's anything else I can do to help get this pull request merged, please let me know :-)

Liryna:
@phobeus I will find some time next week to take a look! :)

samples/dokan_mirror/mirror.c
```diff
@@ -1549,7 +1549,8 @@ void ShowUsage() {
           "  /f User mode Lock\t\t\t\t Enable Lockfile/Unlockfile operations. Otherwise Dokan will take care of it.\n"
           "  /e Disable OpLocks\t\t\t\t Disable OpLocks kernel operations. Otherwise Dokan will take care of it.\n"
           "  /i (Timeout in Milliseconds ex. /i 30000)\t Timeout until a running operation is aborted and the device is unmounted.\n"
-          "  /z Enabled FCB GC. Might speed up on env with filter drivers (Anti-virus) slowing down the system.\n\n"
+          "  /z Enabled FCB GC. Might speed up on env with filter drivers (Anti-virus) slowing down the system.\n"
+          "  /x (network unmount)\t\t\t Allows unmounting network drive from file explorer\n\n"
```
Liryna, line 1553: Would you also be able to add the option in memfs sample ?
https://github.com/dokan-dev/dokany/blob/master/samples/dokan_memfs/main.cpp#L42

Liryna (changes requested):
Awesome changes! Thanks! Work on my side. Could you rebase all the commit into one after adding the memfs flag ?

Liryna:
Hi @phobeus ,

Thank you very much for adding the feature! Everything looks good and can be merged now 👍 
I will try to plan a dokan release in the next weeks (if I do find time).

PR #877, opened 2020-03-29
Modify compilation options: from MD to MT

hefengliang:
I modified all MT options in debug and release modes except the sys project, and all modes compiled successfully.
I did some simple tests on MT's dokan1.dll in win7 64-bit system, but I did not test it on all platforms.
To ensure the reliability of the dokan release package, you can repackage it and remove the vs2019_redist dependency. I integrated it into my application and tested it in a variety of system environments for our users.


Liryna:
Hi @hefengliang ,

Seems like you did the change in the same branch of your previous PR that makes the PR including both commit. Could you redo the PR with only the MT change ? Also change the installer to remove redist dependency and check ?

PR #900, opened 2020-05-30
Description:Fix up dokanfuse mount error when chinese mount path.

FaerieLee:
Fixes # .issue 893 Fix up dokanfuse mount error when chinese mount path.

### Checklist

-
-
-

### DIFF
--- a/dokan_fuse/include/utils.h
+++ b/dokan_fuse/include/utils.h
@@ -11,7 +11,7 @@
 extern "C" {
 #endif
 void utf8_to_wchar_buf_old(const char *src, wchar_t *res, int maxlen);
-void utf8_to_wchar_buf(const char *src, wchar_t *res, int maxlen);
+size_t utf8_to_wchar_buf(const char *src, wchar_t *res, int maxlen);
 
 FILETIME unixTimeToFiletime(time_t t);
 time_t filetimeToUnixTime(const FILETIME *ft);
--- a/dokan_fuse/src/dokanfuse.cpp
+++ b/dokan_fuse/src/dokanfuse.cpp
@@ -515,7 +515,9 @@ int do_fuse_loop(struct fuse *fs, bool mt) {
   }
 
   wchar_t mount[MAX_PATH + 1];
-  mbstowcs(mount, fs->ch->mountpoint.c_str(), MAX_PATH);
+  if (utf8_to_wchar_buf(fs->ch->mountpoint.c_str(), mount, MAX_PATH) == static_cast<size_t>(-1)) {
+    return -1;
+  }
 
   dokanOptions->Version = DOKAN_VERSION;
   dokanOptions->MountPoint = mount;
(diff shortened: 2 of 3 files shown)


Liryna:
Thank you @FaerieLee for the PR.
Why not directly using `utf8_to_wchar_buf` instead of `utf8_to_unicode` ? It was not working ?

FaerieLee:
No, I did not try to use utf8_to_wchar_buf, because it's very low probability that our file system would crash in other scenarios when running in the function , but I couldn't find the reason. So if the mount path of user occasionally trigger the crash, it's a bad experience. So I use  utf8_to_unicode instead of utf8_to_wchar_buf. Thank you for your comment.

FaerieLee:
I reproduce the failure. I find that it's our concurrent problem which result to  utf8_to_wchar_buf crash. I think that utf8_to_wchar_buf can fix up chinese mount path problem, but I need to try to validate

FaerieLee:
I have a question. How can I know the result of calling utf8_to_wchar_buf  ? If utf8_to_wchar_buf  encounts some unexpected exceptions. I don't realize it, And if the funtion which calling utf8_to_wchar_buf  cotinues, the result would be unexpected.

Liryna:
No unfortunately `utf8_to_wchar_buf` do not return any error even if it could since it calls `convert_char` that does return -1 on failure. if you are interested to change the prototype, you can 👍

FaerieLee:
I fix up this question by using  utf8_to_wchar_buf, Please review.

Liryna:
`convert_char` return -1 in failure but you made `utf8_to_wchar_buf` return 0 in case of failure 
https://github.com/dokan-dev/dokany/blob/08502afa1de0b32d8a59d090bfa73ecdd12374ab/dokan_fuse/src/utils.cpp#L144
they should both have the same type of failure otherwise your check would not work

FaerieLee:
The return type of conver_char is size_t, which is unsigned type. Although it return -1,  I determine nothing by this value, because it's always greater or equals than 0. So I made utf8_to_wchar_buf  return 0 in case of failure.

Liryna:
What happen with you condition == 0 of convert return - 1?

FaerieLee:
I didn't understand your idea until now. your idea is that conver_char return - 1, although return value is converted into a unsigned value, I can check failue by this very large unsigned number. If so, I think that utf8_to_wchar_buf should return -1.

FaerieLee:
fix up utf8_to_wchar_buf return -1. please review

Liryna:
Thank you @FaerieLee for the change. The CI has reported different warning during the build
https://ci.appveyor.com/project/Maxhy/dokany/builds/33432130/job/iareml00ux4bp0ee#L2287
Could you take a look at it ?

FaerieLee:
OK, I have fixed up the warning. Please review it.

Liryna:
Thanks for the quick changes and for your contribution! @FaerieLee very appreciated

PR #835, opened 2019-10-09
Kernel - Added support for FileIdExtdBothDirectoryInformation

xgcssch:
The support for FileIdExtdBothDirectoryInformation is required when the target is mapped as a volume into docker containers.

### Checklist

- [x ] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ n/a] I have made corresponding changes to the existing documentation
- [ x] My changes generate no new warnings
- [ x] I have updated the change log (Add/Change/Fix)
- [ n/a] I have cleaned up the commit history (use rebase and squash)

### Changes proposed in this pull request:

Problem:
When a dokan mountpoint is used to map a volume into a Docker container, there is a problem changing to a subdirectory below the mountpoint.

Steps to reproduce
1. A directory `c:\mytarget` with subdirectories `A` and `B`
2.  docker must be installed
3. `mirror.exe /r c:\mytarget /l c:\docker\volume /w /d /s`
4. `docker run -ti -v c:\docker\volume:c:\targetinside mcr.microsoft.com/windows/servercore:1903`
5. Inside the container cd \targetinside
6. `cd B`
7. The new subdirectory is now `A`
The debug output shows:

```
...
###GetFileInfo 0031
GetFileInfo : c:\mytarget\
        GetFileInformationByHandle success, file size = 0
FILE ATTRIBUTE  = 48
        result =  0
        FileInternalInformation
        DispatchQueryInformation result =  0
not suported type 63
not suported type 63
###FindFiles 0031
FindFiles : c:\mytarget\
        FindFiles return 2 entries in c:\mytarget\*

index from 0
FileMatch? : A (*,0,0)
  =>return single entry
index to 1
###GetFileInfo 0029
...
```

63 is `FileIdExtdBothDirectoryInformation`

This pull request adds support for this event.

The correct subdirectory can be selected after applying the fix.

Issue #834, opened 2019-10-08
Dokan not working in Container/Windowsserver Core


xgcssch:
# Environment
 * Windows version: 1809 (17736.737)
 * Processor architecture: x64
 * Dokany version: 1.3.0
 * Library type (Dokany/FUSE): Dokany
 
# Check List
  - [x] I checked my issue [doesn't exist yet]
  - [n/a] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
  - [x] I can always reproduce the issue with the provided description below.
  - [x] I have updated Dokany to the latest version and have reboot my computer after.
  - [x] I tested one of the last snapshot from appveyor CI

# Description
I tried to install dokan in a Windows Server Core container running with docker.
The container base image is Windows Server Core 2016 / 1809.
The installation completes successfully, but the driver is not accessible:

```
"c:\Program Files\Dokan\DokanLibrary-1.3.0\dokanctl.exe" /v
Driver path: 'C:\Windows\system32\drivers\dokan1.sys'
dokanctl : Jul 24 2019 14:10:25
Dokan version : 130
Dokan Error: Failed to open \\.\Dokan_1 with code 2
Dokan driver version : 0x0
```
Neither Application Eventlog nor System shows any hints.

Is Dokan expected to run on Windows Server core?
Is Dokan expected to run on Windows Server core running inside a docker container?

Steps to reproduce:
1. Install [Docker](https://docs.docker.com/docker-for-windows/install/)
2. Create an empty directory and put the following Dockerfile into that directory:
[Dockerfile.txt](https://github.com/dokan-dev/dokany/files/3702570/Dockerfile.txt)
3. Rename Dockerfile.txt to Dockerfile (Github does not allow files without extension)
4. Create a subdirectory `Data` and place `VC_redist.x64.exe`, `Dokan_x64.msi` and `mirror.exe` there
5. Open an admin shell, change to the directory created in step 2., and issue `docker build -t windowscoredokan .`
6. Start the freshly build container with `docker run -ti windowscoredokan`
7. After the container has come up with a commandprompt: `"c:\Program Files\Dokan\DokanLibrary-1.3.0\dokanctl.exe" /v` shows the problem.

Thanks much
Sönke

Liryna:
@xgcssch Hi,

Windows Server Core does not seems to be able to register the driver during install. Can you try to run this command :
pnputil –i –a dokan1.inf
[...]

PR #933, opened 2020-10-08
Return STATUS_INVALID_PARAMETER where appropriate

### LINKED ISSUE #932: WIP: Can't list directories in WSL2
# Note the WIP tag, I gotta grab some more logs and stuff but I've done the analysis and will have a PR incoming shortly
just getting tired of losing my progress every time i reboot

# Environment
 * Windows version: 10.0.19041.508
 * Processor architecture: x86_64
 * Dokany version: 1.4.0.1000
 
# Check List

# Description

A Dokan-based filesystem mounted into a WSL2 instance returns ENOSYS ("Function not implemented") when attempting to list the contents of a directory.

wsl

# Logs
[mirror.log](https://github.com/dokan-dev/dokany/files/5350489/mirror.log)

Please attach in separate files: mirror [output](https://github.com/dokan-dev/dokany/wiki/How-to-Debug-Dokan#mirrorfs), library logs and kernel [logs](https://github.com/dokan-dev/dokany/wiki/How-to-Debug-Dokan#get-logs).
In case of BSOD, please attach minidump or dump [analyze output](https://github.com/dokan
-dev/dokany/wiki/How-to-Debug-Dokan#crash-report-bsod).

reidrankin:
### Checklist

- Returns `STATUS_INVALID_PARAMETER` rather than `STATUS_NOT_IMPLEMENTED` to report the use of an unimplemented information class value with an otherwise implemented IRP type. (Fixes #932.)
- Minor typo edits.
- Minor cleanup in `DokanSetAllocationInformation` to normalize style of code.


Liryna:
Thanks @reidrankin for the PR and the extra changes 🚀 

Changing the default return value in `dokan/volume.c` is fine but it would be better to directly return this status from the kernel to avoid sending the request to the userland for cases
https://github.com/dokan-dev/dokany/blob/91e87827bd76ed8f048dde5bd9e8c7b958f4de3a/dokan/directory.c#L542-L548 that we know that are not supported.
I know that not the case yet but some projects do not directly use the library code so fixing in the driver would also be fixing the issue for them.
Could you add all the previous list of supported infoclass in the kernel here and return the status in the default behaviour ? https://github.com/dokan-dev/dokany/blob/f7161d7f760f8bf3597c0597066192363b4d470c/sys/directory.c#L122-L145
If you could this for the other cases it would be great 👍 I know that might be too much to ask but it would definitely solve the issue.

reidrankin:
I don't think it's necessarily a good idea to limit the range of info classes that the user-mode code can choose to handle. See for example #835, where a new class was implemented without touching the kernel code. So instead I went on a journey trying to figure out how I could intercept `STATUS_NOT_IMPLEMENTED` responses once the user-mode code had had a chance to look at the request and decided it couldn't. I thought about putting it in `DokanCompleteDirectoryControl`, but I don't think this is an `IRP_MJ_DIRECTORY_CONTROL`-specific issue; I ended up putting code in `DokanCompleteIrpRequest` to replace `STATUS_NOT_IMPLEMENTED` with `STATUS_NOT_SUPPORTED` instead.

My reasoning is that Dokan clients use `STATUS_NOT_IMPLEMENTED` to signal that they have not provided an implementation, and that a default behavior should be used instead. No matter what the requested operation or where the implementation might be provided, though, it's certainly got to happen before calling `DokanCompleteIrpRequest`. Therefore, we can confidently state that any function not implemented by this point is not supported by the device at all.

I used `STATUS_NOT_SUPPORTED` instead of `STATUS_INVALID_PARAMETER` in `DokanCompleteIrpRequest` because the only thing that can really decide whether the request failed because a parameter value was bad is the user-mode code. Once the driver hears about it, it can't really be sure if a `STATUS_NOT_IMPLEMENTED` result was caused because of an unsupported info class parameter or something different entirely. For this reason, I've used `STATUS_NOT_SUPPORTED`, because it seems like a more accurate description of what we can be sure of inside `DokanCompleteIrpRequest`, and because it's one of the fallback-triggering codes recognized by WSL2's Plan 9 server.

I've left the library changes that return `STATUS_INVALID_PARAMETER` in response to an unsupported info class in place to provide a more-specific description of these cases (and also works with WSL2).

Finally, I've added support for `FileIdExtdDirectoryInformation` -- I realized we had code for `FileIdExtdBothDirectoryInformation`, and it seemed quite odd to skip the simpler one. This, of course, is the info class that WSL2 first requests, so this will fix the problem in the first instance.

In fact, because the WSL2 Plan 9 server actually comes from the Windows Containers infrastructure, I would bet that issue #834 is caused by a similar setup that would have fallen back to other info classes if it had gotten a slightly different error code. I know that the virtual DVD drive driver doesn't support `FileIdExtdDirectoryInformation`, and I doubt it handles `FileIdExtdBothDirectoryInformation` either. I'm pretty sure those work in containers, though, which I'd expect to be because of a similar fallback mechanism to the one at issue in #932.

reidrankin:
OK, so now that the stupid thing works, let's talk about whether this approach is a good idea or not.

(And I love the Appveyor setup's automagic Driver Verifier test suite, though I love the hour-and-a-half CI pipeline somewhat less.)

Liryna:
Glad you like the CI :)! I also agree that it is pretty slow... Making the test running side by side would be a good start but thats a premium feature :(

Awesome that you added the support for this infoclass! Have you been able to build and test it against WSL 2?

You made a good point of not restricting the infoclass sent by the kernel and for the same reason I would prefer to not override the status sent by the library in the kernel. If other libraries want to get this fix, they will need to import it themselves and that's fine.
Therefore would you be able to just remove the last commit of your PR?

reidrankin:
I've tested the AppVeyor build for 9c62a45, and the new infoclass looks like it works as intended with WSL. I'll reset and force-push this branch back to 4a5078e.

Liryna:
Really by thanks to you @reidrankin for taking your valuable time on this! 🏆 
I might have been long to deal with this PR but if later you will have other changes to propose, I will ensure that I review them more quickly 👍

PR #958, opened 2021-01-10
remove BSF_NOTIMEOUTIFNOTHUNG flag in BroadcastSystemMessage

### LINKED ISSUE #957: DokanBroadcastLink stuck in BroadcastSystemMessage causes timeout
Feature request can skip this form. Bug report **must** complete it. `Check List` must be 100% match or it will be automatically closed without further discussion. Please remove this line.

# Environment
 * Windows version:  Windows 10 build 19042.685
 * Processor architecture:  Intel (i7-9750H)
 * Dokany version:  1.4.0.1000
 * Library type (Dokany/FUSE): Dokany (mirror) 
 
# Check List
  - [ ] I tested one of the last [snapshot](https://github.com/dokan-dev/dokany/wiki/Build#user-snapshot) from appveyor CI 
  
(built from latest github source instead)

On my new notebook, if I mounted an fs (either cppcryptfs or Dokany mirror) using a drive letter, it would get timed out and unmounted by the driver after a short time.

This happend only when I used a drive letter.  It did not happen if I mounted to an empty directory.

I debugged it, and what I found was that after mounting to a drive letter, and before the keepalive setup is done, dokan1.dll calls BroadcastSystemMessage() to inform other apps about the new drive letter.

This call to BroadcasstSystemMessage() is getting stuck either forever or at least long enough for the kernel driver to timeout the fs and unmount it.

If I remove the BSF_NOTIMEOUTIFNOTHUNG flag in the call to BroadcastSystemMessage(), then the call to  BroadcastSystemMessage() returns after about 7 seconds, the keepalive code runs, and everything is ok.

It looks to me that some (probably pre-installed) software on my new notebook is getting the broadcast and not responding to it but also not appearing to Windows to be hung.

bailey27:
Fixes #957.

### Checklist

- [] I have commented my code, particularly in hard-to-understand areas

I think it isn't a significant enough change for updating the change log or even a comment.

I think a rebase/squash isn't needed.  I did only one commit.

Using BSF_NOTIMEOUTIFNOTHUNG in the BroadcastSystemMessage that is done in DokanBroadcastLink after mounting to a drive letter could cause the BroadcastSystemMessage call to block long enough to cause the driver to timeout the fs.  This flag is not necessary so it was removed.
-
-
-

### DIFF
--- a/dokan/mount.c
+++ b/dokan/mount.c
@@ -532,7 +532,7 @@ void DokanBroadcastLink(WCHAR cLetter, BOOL bRemoved, BOOL safe) {
   params.dbcv_flags = 0;
 
   if (BroadcastSystemMessage(
-          BSF_NOHANG | BSF_FORCEIFHUNG | BSF_NOTIMEOUTIFNOTHUNG, &receipients,
+          BSF_NOHANG | BSF_FORCEIFHUNG, &receipients,
           WM_DEVICECHANGE, device_event, (LPARAM)&params) <= 0) {
 
     DbgPrint("DokanBroadcastLink: BroadcastSystemMessage failed - %d\n",


Liryna:
Thank you for the quick pull request! Very appreciated.

bailey27:
You're welcome!

PR #970, opened 2021-02-05
SYS - Check the return code when deleting the file view

zqlustc:
When the deletion of the underlying user-mode file system fails, the file view will still be deleted. File still exists after refreshing.

Fixes # 969

### Checklist

- When user delete the file failed, it will not remove the file view in the window.

### DIFF
--- a/sys/cleanup.c
+++ b/sys/cleanup.c
@@ -210,7 +210,7 @@ VOID DokanCompleteCleanup(__in PIRP_ENTRY IrpEntry,
                             FILE_ACTION_MODIFIED);
   }
 
-  if (DokanFCBFlagsIsSet(fcb, DOKAN_DELETE_ON_CLOSE)) {
+  if (DokanFCBFlagsIsSet(fcb, DOKAN_DELETE_ON_CLOSE) && status == STATUS_SUCCESS) {
     if (DokanFCBFlagsIsSet(fcb, DOKAN_FILE_DIRECTORY)) {
       DokanNotifyReportChange(fcb, FILE_NOTIFY_CHANGE_DIR_NAME,
                               FILE_ACTION_REMOVED);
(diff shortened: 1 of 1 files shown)


Liryna:
@zqlustc The status is always SUCCESS from https://github.com/dokan-dev/dokany/blob/master/dokan/cleanup.c#L37
This PR will not solve the issue you are facing.
Could you open an issue and explain your issue there ?

zqlustc:
> Could you open an iss
Hi,
   I am sorry for not making it clear.
   Now I have a file in the dokan mount point, for example, named test.txt. Right-click to delete the test1 file it will delete test.txt with no prompt, but it maybe failed in the dokan fuse. After I refesh the parent dir of test.txt. It appears again after the directory is refreshed.

Liryna:
@zqlustc Per microsoft documentation, cleanup cannot fail. If delete must fail, it must be during createfile or set delete that everything need to put in place for the delete to work during cleanup.

zqlustc:
Hi,
Thanks, I get it.
Now I have two problems when build teh dokan.

- [1] After I execute '.\scripts\build_installer.ps1', it occur the following errors
C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\km\wdm.h(18870,32): error C4013: “_DataSynchronizationBarrier” not defined； (source file: access.c) [C:\xxxx\dokany-1.4
.1.1000\dokany-1.4.1.1000\sys\sys.vcxproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\km\wdm.h(18870,32): error C4013: “_DataSynchronizationBarrier”not deifned；(source file: close.c) [C:\xxxx\dokany-1.4.
1.1000\dokany-1.4.1.1000\sys\sys.vcxproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\km\wdm.h(19237,22): error C4013: “_MoveFromCoprocessor”not defiend; (source file: access.c) [C:\xxxx\dokany-1.4.1.1000

- [2] We can add a parameter to build x64 only with '.\scripts\build_installer.ps1'. ARM,ARM64,... is useless for me.

Liryna:
Glad you have been able to find a way to get around your cleanup issue.

zqlustc:
Hi,
Sorry to send email to you for help several times. You Are very friendly and outstanding. Thanks very much.


Now I have a problem.&nbsp;
Could we build the x64 setup only with build_installer.ps1? In other word, could I add some parameters to build the x64 setup, not all arch?
Thanks.



---Original---

zqlustc:
Hi,
I have find the method to build x64 setup only.
Thanks.



---Original---

Issue #355, opened 2016-09-27
Mount Manager option show folder as Mounted Volume type


g-fusion:
Feature request can skip this form. Bug report **must** complete it. `Check List` must be 100% match or it will be automatically closed without further discussion. Please remove this line.
# Environment

Windows version: Win 10
Processor architecture: x64
Dokany version: 1.0.0 RC4
Library type (Dokany/FUSE): Dokany
# Check List
- [ x] I checked my issue [doesn't exist yet](https://github.com/dokan-dev/dokany/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [x ] My issue is valid with mirror default sample and not specific to my user-mode driver implementation
- [x ] I can always reproduce the issue with the provided description below.
- [x ] I have updated Dokany to the latest version (https://github.com/dokan-dev/dokany/releases) and have reboot my computer after.
# Description

When using Mount Manager and check Properties of a Folder you do NOT have Size, Size on disk, Contains props.
# Logs

Screenshot attached 
![mountmanager](https://cloud.githubusercontent.com/assets/20126958/18870620/2b247d3e-84ba-11e6-9c92-27b4ffa98902.png)

Rondom:
Is this really a bug?
I can reproduce this on Windows 2016, but the information can be seen by clicking the "Properties"-button next to "Type: Mounted Volume". I think this is just how Windows-Explorer chooses to display the information and nothing that can be "fixed" from the Dokan side.
![win2016_mirror](https://cloud.githubusercontent.com/assets/125913/19746974/3bba44c2-9bd9-11e6-8aaa-69e01d878514.png)

Rondom:
Also, at least on Win2016 it does not matter whether mount manager is used. I get the same behaviour regardless of whether I use the `/o`-option or not.

Liryna:
@Rondom Without setting mount folder and simply with mount manager: `mirror.exe /r C:\ /l m /o`
All subfolders like M:\Users are shown as `Mounted Volume` and there is no size on the folder property. Seems like explorers think that each folder is a device mounted ?

This issue should be on our side otherwise `C:\` should show the same behavior (it also uses mount manager for mounting).

Rondom:
Ah, I see. The behaviour only occurs when

* mounting as a drive letter
* using mount manager

It does not occur for me when mounted as a directory junction (and still using the mount manager).

PR #1038, opened 2021-12-25
FUSE - Fix compare size_t to zero in fuse utils

ATRiiX:
We've got an infinite loop problem in 'convert_char' function when the get_func fail, because the return value in size_t would never less than zero.

### Checklist

- Fix the infinite loop problem by changing several function's return type from size_t to int.
-

### DIFF
--- a/dokan_fuse/include/utils.h
+++ b/dokan_fuse/include/utils.h
@@ -11,7 +11,7 @@
 extern "C" {
 #endif
 void utf8_to_wchar_buf_old(const char *src, wchar_t *res, int maxlen);
-size_t utf8_to_wchar_buf(const char *src, wchar_t *res, int maxlen);
+int utf8_to_wchar_buf(const char *src, wchar_t *res, int maxlen);
 
 FILETIME unixTimeToFiletime(time_t t);
 time_t filetimeToUnixTime(const FILETIME *ft);
--- a/dokan_fuse/src/dokanfuse.cpp
+++ b/dokan_fuse/src/dokanfuse.cpp
@@ -519,7 +519,7 @@ int do_fuse_loop(struct fuse *fs, bool mt) {
   }
 
   wchar_t mount[MAX_PATH + 1];
-  if (utf8_to_wchar_buf(fs->ch->mountpoint.c_str(), mount, MAX_PATH) == static_cast<size_t>(-1)) {
+  if (utf8_to_wchar_buf(fs->ch->mountpoint.c_str(), mount, MAX_PATH) == -1) {
     free(dokanOptions);
     return -1;
   }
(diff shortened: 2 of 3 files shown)


Liryna:
Hi @ATRiiX ,

Thank you for the pull request.
Would you have a sample of input to reproduce the `utf8_to_wchar_buf` loop ?
Using -1 with size_t isn't the most pretty code but it should work

ATRiiX:
Hi @Liryna 

This problem can only be reproduced on a spcific test laptop, and the behavior is that the explorer freeze until it fails.
I dump the process to find the `il` variable was added and much bigger than the `src_len`, but the input seems a normal string contains several Chinise charactors. I copyed it but cannot reproduce it on my own machine.

I think if the input can meet the `utf8_lengths[p[0]] == 0` condition, then return to the `readed < 0` judgment in `convert_char` would never be enter before. After fix it, I found the problem on that laptop was solved.
And these -1 are just some more cleanups.

ATRiiX:
If the input is not a valid utf-8 string, for example, a GBK string in chinese, the function 'utf8_to_wchar_buf' may fail. I have tried a chinese charactor '测', as it's {0xb2, 0xe2} in GBK char array, can trigger the loop problem.
FYI.

Liryna:
Thanks @ATRiiX for the details! Shouldn't we just stop using those custom functions that are error prone and simply use https://stackoverflow.com/a/14809553 ?

ATRiiX:
That's a good idea. Using the standard library function would be more simple and clean.
I can spare some time to write and test it. :)

ATRiiX:
Hi @Liryna , I have update a version using `MultiByteToWideChar` and `WideCharToMultiByte` to convert between utf8 and wchar. And for checking the invalid input of utf8 string, I add a check of `0xfffd` as documented in msdn.
I have test it on our application, and the readdir is working correctly.

Liryna:
Thank you very much @ATRiiX for the contribution and for being so quick! Very appreciated 🏆

Liryna:
@ATRiiX https://github.com/ATRiiX/dokany/commit/3193316abaf4f40eb9ef9a93073a32360a12b22a This looks interesting, do you have a repro or described how it happened ?

ATRiiX:
@Liryna 🍺Happy 2022! Glad to hear you are interested to this patch. I just ported it to the master branch, but I donnot have the environment for testing the self-compiled driver recently. So I have not send the pr yet.
This is to fix issue #355 , as I tested on v1.3.0.1000 earlier last year. The problem is caused by the different behavior of IOCTL_MOUNTDEV_QUERY_DEVICE_NAME on sub directories under windows local drive and dokan drive. I use the windows process monitor to traceing the syscalls of opening the properties, and find that the local dirs ioctl return STATUS_INVALID_PARAMETER where the dokan sub dir return SUCCESS. And I added the original IsVolumeOpen check on this ioctl, but it's not working. Finally found the FsContext can be null by adding log there, so I modify the original IsVolumeOpen. (As far as I remember)
I may still need some days ready to test it after the holiday, or if you can pick it and help testing the properties issue.

Liryna:
Happy 2022 you too! 🥳 Oh I see. Thanks for the detailed response!
I believe it can wait so feel free to ping me when you are able to take a look at it again.

PR #1080, opened 2022-03-21
converting file format to UTF-8 that can be compiled by llvm

gmh5225:
Converting file format to UTF-8 that can be compiled by llvm.

## How to testing?
You can use llvm-msvc to compile it [[link]](https://github.com/NewWorldComingSoon/llvm-msvc-build)


Liryna:
Hi @gmh5225 ,
Looks like file is already UTF-8. What is the issue you are trying to solve ?

gmh5225:
> Hi @gmh5225 , Looks like file is already UTF-8. What is the issue you are trying to solve ?

The original file format is UTF-16 (LE).
1>util\log.c(492,10): fatal error : UTF-16 (LE) byte order mark detected in 'util/ioctl.inc', but encoding is not supported

Liryna:
Thanks @gmh5225 . As Github website is not correctly showing the diff, I preferred to apply the conversion myself.
https://github.com/dokan-dev/dokany/commit/0e2c835b0c9780486f0282615d1bf975ea5d579f
Let me know if there is an issue and thanks for the contribution!

gmh5225:
> 

OK. No problem.

PR #1141, opened 2023-03-31
align buffer at 64-byte boundary so that memcpy can take advantage of…

qiqian:
memcpy can take advantage of erms or avx instructions on moden cpus
these instructions run much faster (up to 20% in my test case) on 64-byte aligned address

### DIFF
--- a/dokan/dokan.c
+++ b/dokan/dokan.c
@@ -1366,8 +1366,8 @@ VOID DOKANAPI DokanMapKernelToUserCreateFileFlags(
 }
 
 VOID DOKANAPI DokanInit() {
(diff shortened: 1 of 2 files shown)


Liryna:
Hi @qiqian ,

This is a great change but unfortunately it will break the library - kernel API which means releasing a major version.
If that's fine for you, I will keep this change and merge it when we plan a new major version!

qiqian:
Hi, it's ok for me, I have made a local workaround to speedup memcpy, just wish it to be merged in the master branch, so I don't have to keep maintaining a custom branch :)

PR #1236, opened 2024-07-28
Fixes for multi-user session issues

LTRData:
Ref discussion: https://github.com/dokan-dev/dokany/discussions/1235

### Checklist

- [X] My code follows the style guidelines of this project


LTRData:
I have implemented fixes for issues discussed https://github.com/dokan-dev/dokany/discussions/1235, including calling `FsRtlRegisterUncProvider` and `FsRtlDeregisterUncProvider` directly from requestor thread instead of a system thread.

I hope this looks good. I have ran lots of successful tests in multi-user environments now to verify that everything works as expected.

sys/init.c
```diff
@@ -757,7 +776,6 @@ DokanCreateGlobalDiskDevice(__in PDRIVER_OBJECT DriverObject,
   return STATUS_SUCCESS;
 }
 
-KSTART_ROUTINE DokanRegisterUncProvider;
 VOID DokanRegisterUncProvider(__in PVOID pDcb) {
```
Liryna, line 779: * This can now directly be typed as `PDokanDCB Dcb` and we can remove https://github.com/dokan-dev/dokany/blob/6a412d7c21bc36db41a1d5a23f6e192e3ae4890c/sys/init.c#L778
* The `PsTerminateSystemThread` call can also be removed
* Same for the `DokanDeregisterUncProvider` function

sys/init.c
```diff
@@ -577,28 +577,49 @@ DokanGetMountPointList(__in PREQUEST_CONTEXT RequestContext) {
...
+  BOOLEAN isAdmin =
+      SeSinglePrivilegeCheck(RtlConvertLongToLuid(SE_IMPERSONATE_PRIVILEGE),
+                             RequestContext->Irp->RequestorMode);
+
+  ULONG sessionId;
+  if (!NT_SUCCESS(IoGetRequestorSessionId(RequestContext->Irp, &sessionId))) {
+    sessionId = 0;
+  }
+
+  status = STATUS_SUCCESS;
```
Liryna, line 589: That's a valid change but maybe then directly set `STATUS_SUCCESS` at the declaration of the variable ? (instead of `STATUS_INVALID_PARAMETER`)

dokan_np/dokannp.c
```diff
@@ -272,10 +272,17 @@ DWORD APIENTRY NPGetConnection(__in LPWSTR LocalName, __out LPWSTR RemoteName,
     return WN_NOT_CONNECTED;
   }
 
+  DWORD currentSessionId;
+  if (!ProcessIdToSessionId(GetCurrentProcessId(), &currentSessionId))
+  {
```
Liryna, line 277: nit: Move the `{` at the end of the above line

sys/init.c
```diff
@@ -577,28 +577,49 @@ DokanGetMountPointList(__in PREQUEST_CONTEXT RequestContext) {
...
+  for (listEntry = RequestContext->DokanGlobal->MountPointList.Flink;
+       listEntry != &RequestContext->DokanGlobal->MountPointList;
+       listEntry = listEntry->Flink, ++i) {
+
+    mountEntry = CONTAINING_RECORD(listEntry, MOUNT_ENTRY, ListEntry);
+    ExAcquireResourceSharedLite(&mountEntry->Resource, TRUE);
+
+    __try {
+      // For non-admins, skip drives that are not in same session as
+      // requestor
```
Liryna, line 606: nit: This does not fit on one line ?

Liryna (commented):
Thank you @LTRData for the quick change! Added a few comments but overall LGTM!

Liryna:
Interesting the test failed for UNC on Appveyor image 2015 but worked on the 2017
https://ci.appveyor.com/project/Maxhy/dokany/builds/50299852/job/d97jyx1t9mdu93sr

sys/init.c
LTRData: Good point. Old habits to always line break at 79 characters, but that is pretty much pointless these days. I'll change it.

LTRData:
Fixed the issues. As a sidenote, for another discussion, I believe the other `RunAsSystem` calls can be changed to specifying global namespace as part of path to `IoCreateSymbolicLink` instead.

LTRData:
Test still fail. Even the tests that are marked as successful (such as 2017) are actually failing. It shows in the logs, although they for some reason summarize the failures in a different way.

So, I'll revert the `RunAsSystem` changes for now and try to find out more about how and why that change caused failures!

sys/fscontrol.c
```diff
@@ -952,7 +952,7 @@ NTSTATUS DokanMountVolume(__in PREQUEST_CONTEXT RequestContext) {
   }
 
   if (isNetworkFileSystem) {
-    RunAsSystem(DokanRegisterUncProvider, dcb);
+    DokanRegisterUncProvider(dcb);
```
Liryna, line 955: Could you also revert this ?

Liryna:
Yeah that's surprising 😢 Agree we can look at this later

sys/fscontrol.c
LTRData: Oh sorry, missed that! I will fix!

LTRData:
Tests successful. I think it all looks good now.

dokan_np/dokannp.c
```diff
@@ -272,10 +272,16 @@ DWORD APIENTRY NPGetConnection(__in LPWSTR LocalName, __out LPWSTR RemoteName,
...
+  DWORD currentSessionId;
+  if (!ProcessIdToSessionId(GetCurrentProcessId(), &currentSessionId)) {
+    currentSessionId = 0;
+  }
+
   dosDevice[12] = LocalName[0];
 
   for (unsigned int i = 0; i < nbRead; ++i) {
-    if (wcscmp(dokanMountPointInfo[i].MountPoint, dosDevice) == 0) {
+    if (dokanMountPointInfo[i].SessionId == currentSessionId &&
```
Liryna, line 283: I believe `dokanMountPointInfo[i].SessionId` is only set if `DOKAN_EVENT_CURRENT_SESSION` was used
https://github.com/dokan-dev/dokany/blob/master/sys/event.c#L837
Should we here only check if `dokanMountPointInfo[i].SessionId != -1` ?

sys/init.c
```diff
@@ -571,30 +571,48 @@ PMOUNT_ENTRY FindMountEntryByName(__in PDOKAN_GLOBAL DokanGlobal,
...
+  for (listEntry = RequestContext->DokanGlobal->MountPointList.Flink;
+       listEntry != &RequestContext->DokanGlobal->MountPointList;
+       listEntry = listEntry->Flink, ++i) {
+
+    mountEntry = CONTAINING_RECORD(listEntry, MOUNT_ENTRY, ListEntry);
+    ExAcquireResourceSharedLite(&mountEntry->Resource, TRUE);
+
+    __try {
+      // For non-admins, skip drives that are not in same session as requestor
+      if (!isAdmin && mountEntry->MountControl.SessionId != sessionId) {
```
Liryna, line 604: Same as the other comment, should we return mount point that are not mounted with `DOKAN_EVENT_CURRENT_SESSION` ( sessionId = -1) ?

dokan_np/dokannp.c
LTRData: Very good point. I will change that both in `DokanGetMountPointList` and `NPGetConnection`.

Liryna:
Yeap looks good! Thank @LTRData for the contribution 👍 I will try to make a release ASAP if you need a release

LTRData:
> Yeap looks good! Thank @LTRData for the contribution 👍 I will try to make a release ASAP if you need a release

Thanks! That would be great! 👍

PR #1253, opened 2024-11-28
Fix possible failure to create file/directory

pomaroff:
If there are 2 (or more) concurrent requests to access a file, any 1 request with a CREATE disposition will be rejected with `STATUS_OBJECT_NAME_COLLISION`. The driver should not return this error code based on the fact that there other open requests before calling userland code to confirm that the file handle is actually open and that the file or directory exists.

We also do not need to run file sharing logic for the first of many concurrent requests on the same file.

### DIFF
--- a/sys/create.c
+++ b/sys/create.c
@@ -675,6 +675,7 @@ Return Value:
       }
     }
     if (allocateCcb) {
+      BOOLEAN isAlreadyOpen = FALSE;
       // Allocate an FCB or find one in the open list.
       if (RequestContext->IrpSp->Flags & SL_OPEN_TARGET_DIRECTORY) {
         status = DokanGetParentDir(fileName, &parentDir, &parentDirLength);
@@ -683,9 +684,11 @@ Return Value:
           fileName = NULL;
           __leave;
         }
-        fcb = DokanGetFCB(RequestContext, parentDir, parentDirLength);
+        fcb = DokanGetFCB(RequestContext, parentDir, parentDirLength,
+                          &isAlreadyOpen);
       } else {
-        fcb = DokanGetFCB(RequestContext, fileName, fileNameLength);
+        fcb = DokanGetFCB(RequestContext, fileName, fileNameLength,
+                          &isAlreadyOpen);
       }
       if (fcb == NULL) {
         status = STATUS_INSUFFICIENT_RESOURCES;
@@ -699,7 +702,7 @@ Return Value:
       DOKAN_LOG_FINE_IRP(RequestContext, "Use FCB=%p", fcb);
 
       // Cannot create a file already open
-      if (fcb->FileCount > 1 && disposition == FILE_CREATE) {
+      if (isAlreadyOpen && disposition == FILE_CREATE) {
         status = STATUS_OBJECT_NAME_COLLISION;
         __leave;
       }
--- a/sys/create.c
+++ b/sys/create.c
@@ -697,12 +697,6 @@ Return Value:
                      &fileObject->FileName);
       }
       DOKAN_LOG_FINE_IRP(RequestContext, "Use FCB=%p", fcb);
-
-      // Cannot create a file already open
-      if (fcb->FileCount > 1 && disposition == FILE_CREATE) {
-        status = STATUS_OBJECT_NAME_COLLISION;
-        __leave;
-      }
     }
 
     // Cannot create a directory temporary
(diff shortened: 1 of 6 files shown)


Liryna:
Hi @pomaroff ,

This change seems to make sense! Thank you for providing a fix.

> To be tested.

May I ask how did you discover this ?

pomaroff:
Hi @Liryna,

Apologies for the scant details, I was hoping to use the build from AppVeyor to test the changes, but I've been unable to install the artifact from there. I'm not sure if it's related to the build failures that are currently happening on master? I may need some help getting an installer to verify the fix, or if you're happy to verify it for me then that'd be great.

> May I ask how did you discover this ?

I discovered this issue because I'm trying to use MsBuild for some C# solutions inside a Dokan mirror, and this is failing on some solutions. I found that the CreateFile callback in userland code wasn't getting called after looking at procmon, where I found that dotnet.exe can rarely make concurrent requests to create the same directory which both receive `STATUS_OBJECT_NAME_COLLISION` responses that it is seemingly unable to recover from. See highlighted rows from procmon here:

> image

The only relevant place in the Dokan driver where this status is returned without calling back to the userland code is the area I've been looking at in this PR. I'm not entirely convinced this code is necessary at all because we could handle the concurrent requests in the userland code, but nevertheless, I think the changes I'm proposing here would be sufficient to get things working?

This is relatively easy to reproduce in a simple test. See this C# example below, which for me tends to fail after a few loops.

```csharp
for (var i = 0; i < 100; i++)
{
    var path = Path.Combine(mount.Path, $"Directory{i}");
    var task1 = Task.Run(() => Directory.CreateDirectory(path));
    var task2 = Task.Run(() => Directory.CreateDirectory(path));
    Task.WaitAll(task1, task2);
    Assert.That(Directory.Exists(path), Is.True, () => $"Failed to create Directory{i}");
}
```

pomaroff:
I've managed to install this locally, and whilst I think it has improved it it hasn't fixed all scenarios. I will look closer into this area of the code.

Liryna:
> I'm not entirely convinced this code is necessary at all because we could handle the concurrent requests in the userland code, but nevertheless,

I tend to agree with you. It is such checks that need to be there for correctness and could be in userland or in the driver. Having it in userland means having all filesystem implementation take care of it (count active handles etc).

Thanks for the update @pomaroff . Please let me know if you find the other reason MsBuild fails.

Liryna:
`STATUS_OBJECT_NAME_COLLISION` must be returned when the caller tries to create an item that already exists.
As you perfectly explained it, we use `fcb->FileCount` as an optimization to know that a handle is open (and therefore the item exists) BUT using `fcb->FileCount` is indeed wrong because it is incremented before userland gets the request and confirm whether an item was actually opened successfully (= exist).

We could introduce a new `fcb->OpenHandleCount` that is incremented on create completion success but that seems complex for just this check that can be handled by userland. And anyway userland could fail createfile for X Y Z reasons while the item actually does exist.

What I am trying to say is that we should drop this if and let the userland implementation take care of it (which is the contrary to what I said previously I know 😄  ). Feel free to change your PR to remove this if, I can look to add it to the mirror / memfs samples.

pomaroff:
Thanks Liryna.

Regarding this:
> `STATUS_OBJECT_NAME_COLLISION` must be returned when the caller tries to create an item that already exists.

I think we already cover this in the documentation:[documentation](https://dokan-dev.github.io/dokany-doc/html/struct_d_o_k_a_n___o_p_e_r_a_t_i_o_n_s.html#a40c2f61e1287237f5fd5c2690e795183):
> In case OPEN_ALWAYS & CREATE_ALWAYS are successfully opening an existing file, STATUS_OBJECT_NAME_COLLISION should be returned instead of STATUS_SUCCESS . This will inform Dokan that the file has been opened and not created during the request.

Liryna:
Correct! 😎 good eye

pomaroff:
What do you think of the other change in this PR, which is to move the VCB Unlock until after the FCB is Locked? In doing so, we ensure that:
1. Requests on the same file are processed in the order received.
2. We read `FileCount` before releasing the VCB lock, so that even if there's another request queued on the same file, the first request will bypass any "file sharing" logic.

Liryna:
Holding the Vcb lock does resolve that issue but is usually not safe unless made very carefully. If any of the called function triggers a reentry (it easily happens) that also end up requiring the Vcb lock, we will deadlock.

It is safer to just move the condition to userland.

pomaroff:
Thanks @Liryna. I'm not sure if it would have caused a deadlock, but I've reverted it to be on the safe side as it's not strictly relevant to the root problem I'm trying to solve here. I can always raise another PR in the future if needed. Thanks.

PR #1330, opened 2026-03-20
np : allow same user in a multi-session environment (Windows Server)

lanopk:
Fixes # .

### Checklist

We have modified the Network Provider to allow requests from the same user in a multi-session environment (Windows Server).
This may also be helpful when user sessions change.
Please refer to the issue below.
https://github.com/dokan-dev/dokany/issues/1328#issuecomment-4093119843

-
-
-

### DIFF
--- a/dokan_np/dokannp.c
+++ b/dokan_np/dokannp.c
@@ -33,6 +33,8 @@ with this program. If not, see <http://www.gnu.org/licenses/>.
 #include <strsafe.h>
 #include <windows.h>
 #include <winnetwk.h>
+#include <wtsapi32.h>
+#pragma comment(lib, "wtsapi32.lib")
 
 static VOID DokanDbgPrintW(LPCWSTR format, ...) {
   const WCHAR *outputString;
@@ -55,6 +57,8 @@ static VOID DokanDbgPrintW(LPCWSTR format, ...) {
 }
 
 #define DbgPrintW(format, ...) DokanDbgPrintW(format, __VA_ARGS__)
+
+BOOL IsSameUser(DWORD sessionID1, DWORD sessionID2);
 
 DWORD APIENTRY NPGetCaps(DWORD Index) {
   DWORD rc = 0;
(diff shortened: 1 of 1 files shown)


lanopk:
It failed the verification. What should I do?
You can simply review the modified parts and edit the code yourself.

LTRData:
* FOSSA License Compliance issues detect that we are using WiX Toolset libraries which are under MIT or MS-RL license while Dokany is under LGPL, which violates LGPL compliance. LGPL practically only allows other LGPL licensed libraries. I guess this might be possible to solve by changing the dependency tree such that nothing LGPL licensed depends on WiX but instead the other way around. It is after all not actually used by the Dokany library code. We probably need to look into this.

* Appveyor build fails because git warns about normalizing line endings (LF-CRLF). This could in most cases be ignored at build servers. I'll see if I can modify the yml to add a command to simply stop doing this on the build server.

* Codacy complains that wtsapi32.h is not found. That is really strange. I wonder if something about Windows SDK setup is wrong in the configuration there. But it seems I cannot see any configuration there, @Liryna maybe?

LTRData:
More about appveyor. If I am not mistaken, it should be possible to solve by adding this to `appveyor.yml`:
```
  init:
    - git config --global core.autocrlf false
    - git config --global core.safecrlf false
```

And possibly also add a `.gitattributes` file with:
```
* -text
```

lanopk:
For reference, the SDKs I used for testing were Windows SDK 10.0.19041.0, 10.0.26100.0, etc., and I built it with Visual Studio 2024. (Because I didn't have Visual Studio 2019)

LTRData:
> For reference, the SDKs I used for testing were Windows SDK 10.0.19041.0, 10.0.26100.0, etc., and I built it with Visual Studio 2024. (Because I didn't have Visual Studio 2019)

Yes, this is most probably an issue with configuration on the build server. Not something in your code.

